PR #1, opened 2022-04-14
feat: initial bulletproofs-plus commit

hansieodendaal:
This is an implementation for Bulletproofs+ based on the proof of concept Python implementation at https://github.com/AaronFeickert/pybullet-plus by [Aaron Feickert](https://github.com/AaronFeickert). Bulletproofs+ was originally intended for [Monero](https://www.getmonero.org/2020/12/24/Bulletproofs+-in-Monero.html). 

Featuring `prove()` and `verify()` for:
  - a single proof (i.e. for one Pedersen commitment);
  - an aggregated proof (i.e. for multiple Pedersen commitment by a single party who knows all the secrets);
  - a batch with multiple single proofs (i.e. from different parties without any dependencies);
  - a batch with multiple mixed single and aggregated proofs (i.e. from different parties without any dependencies).

Also featuring:
- _mask_ (i.e. _blinding factor_) recovery for single proofs where a secret _seed_ was provided to create the proof and where the _seed_ is provided when verifying the proof
- **edit:** minimum value promise (offset) as an optional input to creating and verifying the proof

-----------------

**Creation performance benchmark (non `simd_backend`) - _Bulletproofs+ vs. Dalek Bulletproofs:_**

Size | BP mean (ms) | BP median (ms) | BP+ mean (ms) | BP+ median (ms) | Diff Mean (%) | Diff Median (%)
-- | -- | -- | -- | -- | -- | --
1 | 17.153 | 16.294 | 21.408 | 21.358 | 125% | 131%
2 | 31.907 | 31.630 | 42.101 | 41.982 | 132% | 133%
4 | 60.810 | 60.467 | 84.785 | 82.063 | 139% | 136%
8 | 118.530 | 119.180 | 160.020 | 157.310 | 135% | 132%
16 | 240.470 | 240.180 | 309.500 | 308.310 | 129% | 128%
32 | 471.710 | 460.670 | 619.320 | 615.700 | 131% | 134%

image

-----------------

**Verification performance benchmark (non `simd_backend`) - _Bulletproofs+ vs. Dalek Bulletproofs:_**

Size | BP mean (ms) | BP median (ms) | BP+ mean (ms) | BP+ median (ms) | Diff Mean (%) | Diff Median (%)
-- | -- | -- | -- | -- | -- | --
1 | 2.187 | 2.338 | 2.299 | 2.117 | 105% | 91%
2 | 3.754 | 3.758 | 3.786 | 3.507 | 101% | 93%
4 | 6.440 | 6.443 | 6.488 | 5.885 | 101% | 91%
8 | 11.023 | 11.102 | 11.446 | 10.359 | 104% | 93%
16 | 18.940 | 17.573 | 20.269 | 18.600 | 107% | 106%
32 | 36.687 | 33.685 | 39.547 | 35.519 | 108% | 105%

image

-----------------

**Batched verification performance benchmark (with non `simd_backend`) - _Batched vs. Single proof verification_**

Size | BP+ Linear (ms) | BP+ mean (ms) | BP+ median (ms) | Diff (%)
-- | -- | -- | -- | --
1 | 2.249 | 2.249 | 2.089 | 100%
2 | 4.498 | 2.804 | 2.648 | 62%
4 | 8.997 | 4.343 | 3.744 | 48%
8 | 17.994 | 6.098 | 5.700 | 34%
16 | 35.987 | 10.640 | 9.636 | 30%
32 | 71.974 | 18.750 | 17.280 | 26%

image


src/range_proof.rs
```diff
@@ -0,0 +1,555 @@
...
+                    "Vector L length not adequate".to_string(),
+                ));
+            }
+
+            // Helper values
+            let batch_size = commitments.len();
+            let rounds = li.len();
+
+            // Batch weight (may not be equal to a zero valued scalar)
+            let weight = Scalar::random(rng);
```
AaronFeickert, line 386: Can't this function return zero? If so, need to check and loop until a nonzero scalar is returned.

src/range_proof.rs
```diff
@@ -0,0 +1,555 @@
...
+
+            // Start the transcript
+            let mut transcript = Transcript::new(transcript_label.as_bytes());
+            transcript.domain_separator(b"Bulletproof+", b"Range Proof");
+            transcript.validate_and_append_point(b"H", &h_base.compress())?;
+            transcript.validate_and_append_point(b"G", &h_base.compress())?;
```
AaronFeickert, line 392: Needs to use `g_base`.

src/range_proof.rs
```diff
@@ -0,0 +1,555 @@
...
+            }
+            for i in 0..(statements[index].commitments.len()) {
+                transcript.validate_and_append_point(
+                    b"Ci",
+                    &statements[index].commitments[i].compress(),
+                )?;
+            }
+
+            // Reconstruct challenges
+            transcript.validate_and_append_point(b"H", &proof.a)?;
```
AaronFeickert, line 415: Label should be `A`.

src/range_proof.rs
```diff
@@ -0,0 +1,555 @@
...
+            for j in 0..rounds {
+                transcript.validate_and_append_point(b"L", &li[j].compress())?;
+                transcript.validate_and_append_point(b"R", &ri[j].compress())?;
+                let e = transcript.challenge_scalar(b"e");
+                challenges.push(e);
+                challenges_inv.push(e.invert());
+            }
+            transcript.validate_and_append_point(b"A1", &a1.compress())?;
+            transcript.validate_and_append_point(b"B", &b.compress())?;
+            let e = transcript.challenge_scalar(b"e");
```
AaronFeickert, line 430: The challenges must be nonzero. This is exceptionally unlikely, but it's inexpensive to check this.

src/range_proof.rs
```diff
@@ -0,0 +1,555 @@
...
+        transcript.append_u64(b"N", bit_length as u64);
+        transcript.append_u64(b"M", batch_size as u64);
+        for item in hi_base.clone() {
+            transcript.validate_and_append_point(b"Hi", &item.compress())?;
+        }
+        for item in gi_base.clone() {
+            transcript.validate_and_append_point(b"Gi", &item.compress())?;
+        }
+        for item in statement.commitments.clone() {
+            transcript.validate_and_append_point(b"Ci", &item.compress())?;
```
AaronFeickert, line 151: Zero is a valid commitment.

src/range_proof.rs
```diff
@@ -0,0 +1,555 @@
...
+                )?;
+            }
+            for i in 0..statements[index].generators.gi_base().len() {
+                transcript.validate_and_append_point(
+                    b"Gi",
+                    &statements[index].generators.gi_base()[i].compress(),
+                )?;
+            }
+            for i in 0..(statements[index].commitments.len()) {
+                transcript.validate_and_append_point(
```
AaronFeickert, line 408: Zero is a valid commitment.

src/range_proof.rs
```diff
@@ -0,0 +1,555 @@
...
+            let z = transcript.challenge_scalar(b"z");
+            transcript.domain_separator(b"Bulletproof+", b"Inner Product Proof");
+            let mut challenges = vec![];
+            let mut challenges_inv = vec![];
+            for j in 0..rounds {
+                transcript.validate_and_append_point(b"L", &li[j].compress())?;
+                transcript.validate_and_append_point(b"R", &ri[j].compress())?;
+                let e = transcript.challenge_scalar(b"e");
+                challenges.push(e);
+                challenges_inv.push(e.invert());
```
AaronFeickert, line 426: You can replace these individual inversions with a single slice inversion using `Scalar::batch_invert` for improved efficiency. It is _extremely important_ that no element of the input slice be zero in this case.

AaronFeickert (changes requested):

src/range_proof.rs
hansieodendaal: It is initialized with zero (`let mut scalar_bytes = [0u8; 64];`) and then filled with new random data.

src/range_proof.rs
```diff
@@ -0,0 +1,555 @@
...
+                    b"Hi",
+                    &statements[index].generators.hi_base()[i].compress(),
+                )?;
+            }
+            for i in 0..statements[index].generators.gi_base().len() {
+                transcript.validate_and_append_point(
+                    b"Gi",
+                    &statements[index].generators.gi_base()[i].compress(),
+                )?;
+            }
```
AaronFeickert, line 406: Since each proof uses a subset of a common generator set, you could do one of a couple of things to make this more efficient:
1. Leave out these generators, provided they are guaranteed never to change or be under adversarial control. This is a somewhat brittle approach.
2. Precompute domain-separated hashes of each subset (or hardcode them and include a test for their correctness), and use one of these hashes in the transcript instead. This is a safer approach.

src/range_proof.rs
AaronFeickert: Right, but in theory it could return zero (or if the RNG were to fail somehow). It's cheap to check.

hansieodendaal: O second thought, I could not find the check, so adding one.

AaronFeickert: I did find a documentation reference that suggested the underlying byte generator _should_ not return the same output as the (zero) input, but adding a specific check here seems safer and more robust against implementation-specific behavior; it also makes the intent of the protocol more clear.

src/range_proof.rs
AaronFeickert: On further thought and inspection, I think you can safely remove these generator vectors from the transcript, since they can't be externally changed when a statement and proof are presented to the verifier. That being said, it's extremely important that the implementation not change the way these are generated. If this is at all a possibility, you could include a version number in the transcript, and increment it if the generators are ever changed.

I do think the `G` and `H` generators should still be included (even though they are also not specifically presented to the verifier), since this is cheap.

AaronFeickert (approved):
Reviewed changes and confirmed tests pass.

src/lib.rs
```diff
@@ -0,0 +1,47 @@
+// Copyright 2022 The Tari Project
+// SPDX-License-Identifier: BSD-3-Clause
+
+#![cfg_attr(not(debug_assertions), deny(unused_variables))]
```
mikethetike, line 4: These can come out since they are duplicated in `lints.toml`

src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+#![cfg_attr(not(debug_assertions), deny(unused_extern_crates))]
+#![deny(unused_must_use)]
+#![deny(unreachable_patterns)]
+#![deny(unknown_lints)]
+#![recursion_limit = "1024"]
+// #![deny(missing_docs)]
+
+// Some functions have a large amount of dependencies (e.g. services) and historically this warning has lead to
+// bundling of dependencies into a resources struct, which is then overused and is the wrong abstraction
+#![allow(clippy::too_many_arguments)]
```
mikethetike, line 16: This should also be in `lints.toml`

src/range_parameters.rs
```diff
@@ -0,0 +1,100 @@
...
+pub struct RangeParameters {
+    /// Generators needed for aggregating up to `m` range proofs of up to `n` bits each.
+    bp_gens: BulletproofGens,
+    /// The pair of base points for Pedersen commitments.
+    pc_gens: PedersenGens,
+}
+
+impl RangeParameters {
+    /// Initialize a new 'RangeParameters' with sanity checks
+    pub fn init(bit_length: usize, batch_size: usize) -> Result<RangeParameters, ProofError> {
```
mikethetike, line 27:
```suggestion
    pub fn new(bit_length: usize, batch_size: usize) -> Result<Self, ProofError> {
```

src/range_parameters.rs
```diff
@@ -0,0 +1,100 @@
...
+        self.pc_gens.b_base
+    }
+
+    /// Return the protected mask base point
+    pub fn g_base(&self) -> RistrettoPoint {
+        self.pc_gens.b_base_blinding
+    }
+
+    /// Return the protected value bulletproof generators
+    pub fn hi_base(&self) -> Vec<RistrettoPoint> {
```
mikethetike, line 82: Rather return a slice with pointers

src/range_proof.rs
```diff
@@ -0,0 +1,558 @@
+// Copyright 2022 The Tari Project
+// SPDX-License-Identifier: BSD-3-Clause
+
+#![deny(missing_docs)]
```
mikethetike, line 4:
```suggestion

```

src/range_statement.rs
```diff
@@ -0,0 +1,70 @@
...
+impl RangeStatement {
+    /// Initialize a new 'RangeStatement' with sanity checks
+    pub fn init(
+        generators: RangeParameters,
+        commitments: Vec<RistrettoPoint>,
+        minimum_value_promise: Vec<Option<u64>>,
+        seed_nonce: Option<Scalar>,
+    ) -> Result<RangeStatement, ProofError> {
+        if !commitments.len().is_power_of_two() {
+            return Err(ProofError::InternalDataInconsistent(
```
mikethetike, line 37: This is used a lot, perhaps a better name would be `ProofError::InvalidArgument`

src/range_proof.rs
```diff
@@ -0,0 +1,558 @@
...
+            Ok(ri)
+        } else {
+            Err(ProofError::InternalDataInconsistent(
+                "Vector 'R' not assigned yet".to_string(),
+            ))
+        }
+    }
+
+    /// Create a single or aggregated range proof for a single party that knows all the secrets
+    pub fn prove(
```
mikethetike, line 110: Maybe add a docstring describing how the transcript should be created. Should it be a new instance? Under what circumstances can an existing Transcript be used?

src/commitment_opening.rs
```diff
@@ -0,0 +1,31 @@
...
+
+#![deny(missing_docs)]
+
+//! Bulletproof+ commitment opening struct
+
+use curve25519_dalek::scalar::Scalar;
+use zeroize::Zeroize;
+
+/// Commitment openings to be used for Pedersen commitments
+#[derive(Clone, Debug, Zeroize)]
```
mikethetike, line 12: Remember to hide these from debug

src/range_proof.rs
```diff
@@ -0,0 +1,558 @@
...
+        statement: &RangeStatement,
+        witness: RangeWitness,
+    ) -> Result<RangeProof, ProofError> {
+        let batch_size = statement.commitments.len();
+        if witness.openings.len() != batch_size {
+            return Err(ProofError::InvalidLength(
+                "Invalid range statement - commitments and openings do not match!".to_string(),
+            ));
+        }
+        for j in 0..batch_size {
```
mikethetike, line 121: This step seems superfluous. The caller either has valid data or can do this check themselves.

If it must be done, then a single `multiscalar_mult` would be better

src/range_parameters.rs
mikethetike, line 82: This clone is unnecessary, rather return a `&[RistrettoPoint]`

src/range_parameters.rs
```diff
@@ -0,0 +1,100 @@
...
+        let hi_base: Vec<RistrettoPoint> = self
+            .bp_gens
+            .h_iter(self.bit_length(), self.batch_size())
+            .cloned()
+            .collect();
+        hi_base
+    }
+
+    /// Return the protected mask bulletproof generators
+    pub fn gi_base(&self) -> Vec<RistrettoPoint> {
```
mikethetike, line 92: Same

src/range_proof.rs
```diff
@@ -0,0 +1,558 @@
...
+        let bit_length = statement.generators.bit_length();
+
+        // Global generators
+        let h_base = statement.generators.h_base();
+        let g_base = statement.generators.g_base();
+        let hi_base = statement.generators.hi_base();
+        let gi_base = statement.generators.gi_base();
+
+        transcript.domain_separator(b"Bulletproof+", b"Range Proof");
+        transcript.validate_and_append_point(b"H", &h_base.compress())?;
```
mikethetike, line 143: You can compress these once in the generators, and return the compressed data directly

src/range_proof.rs
```diff
@@ -0,0 +1,558 @@
...
+        transcript.validate_and_append_point(b"G", &g_base.compress())?;
+        transcript.append_u64(b"N", bit_length as u64);
+        transcript.append_u64(b"M", batch_size as u64);
+        for item in statement.commitments.clone() {
+            transcript.append_point(b"Ci", &item.compress());
```
mikethetike, line 148: I would suggest storing both the compressed and uncompressed versions of the commitments in the statement. The caller might already have both representations at hand. If they do not, then it can be compressed upfront in the statement

src/range_proof.rs
```diff
@@ -0,0 +1,558 @@
...
+        for item in statement.minimum_value_promises.clone() {
+            if let Some(minimum_value) = item {
+                transcript.append_u64(b"vi - minimum_value", minimum_value);
+            } else {
+                transcript.append_u64(b"vi - minimum_value", 0);
+            }
+        }
+
+        // Set bit arrays
+        let mut a_li = vec![];
```
mikethetike, line 159:
```suggestion
        let mut a_li = Vec::with_capacity();
```

src/range_proof.rs
```diff
@@ -0,0 +1,558 @@
...
+        let a = RistrettoPoint::multiscalar_mul(ai_scalars, ai_points);
+        transcript.validate_and_append_point(b"A", &a.compress())?;
+
+        // Get challenges
+        let y = transcript.challenge_scalar(b"y")?;
+        let z = transcript.challenge_scalar(b"z")?;
+        let z_square = z * z;
+
+        // Compute powers of the challenge
+        let mut y_powers = vec![Scalar::one()];
```
mikethetike, line 205:
```suggestion
        let mut y_powers = Vec::with_capacity(batch_size*bit.......);
         y_powers.push(Scalar::one());
```

mikethetike (commented):
Still busy, but some comments so far

src/commitment_opening.rs
```diff
@@ -0,0 +1,31 @@
+// Copyright 2022 The Tari Project
+// SPDX-License-Identifier: BSD-3-Clause
+
+#![deny(missing_docs)]
```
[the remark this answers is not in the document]
hansieodendaal: Yip, removed those, an artefact of getting all of it done

src/inner_product_round.rs
```diff
@@ -0,0 +1,359 @@
...
+            done: false,
+            ai: a,
+            bi: b,
+            alpha,
+            a1: None,
+            b: None,
+            r1: None,
+            s1: None,
+            d1: None,
+            li: Vec::new(),
```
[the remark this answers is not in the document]
AaronFeickert: Capacity of the inner-product vectors is fixed by the bit length and aggregation factor, so it's known from the start.

src/inner_product_round.rs
```diff
@@ -0,0 +1,359 @@
...
+
+            self.r1 = Some(r + self.ai[0] * e);
+            self.s1 = Some(s + self.bi[0] * e);
+            self.d1 = Some(eta + d * e + self.alpha * e * e);
+
+            return Ok(());
+        };
+
+        n = div_floor_usize(n as f32, 2f32);
+        let a1 = self
```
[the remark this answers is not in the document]
hansieodendaal: I removed the use of `NonEmpty`

src/inner_product_round.rs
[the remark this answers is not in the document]
hansieodendaal: Got it:
``` rust
                li: Vec::with_capacity(n * batch_size + 2),
                ri: Vec::with_capacity(n * batch_size + 2),
```

src/utils.rs
```diff
@@ -0,0 +1,222 @@
...
+            result.push(Scalar::one());
+        } else {
+            result.push(Scalar::zero());
+        }
+    }
+    Ok(result)
+}
+
+/// A convenience function ot calculate division and return the rounded down value
+pub fn div_floor_usize(value: f32, divisor: f32) -> usize {
```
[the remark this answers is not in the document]
hansieodendaal: Yes, this was not needed

src/range_proof.rs
```diff
@@ -178,7 +179,7 @@ impl RangeProof {
             ai_scalars.push(a_ri[i]);
             ai_points.push(hi_base[i]);
         }
-        let a = RistrettoPoint::multiscalar_mul(ai_scalars, ai_points);
+        let a = RistrettoPoint::vartime_multiscalar_mul(ai_scalars, ai_points);
```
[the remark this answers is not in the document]
hansieodendaal: Agreed, maybe it is not a concern. If it turns out to be, we can change it.

src/inner_product_round.rs
```diff
@@ -0,0 +1,320 @@
...
+            li_points.push(hi_base_lo[i]);
+            ri_scalars.push(a2[i] * self.y_powers[n]);
+            ri_points.push(gi_base_lo[i]);
+            ri_scalars.push(b1[i]);
+            ri_points.push(hi_base_hi[i]);
+        }
+        self.li
+            .push(RistrettoPoint::vartime_multiscalar_mul(li_scalars, li_points));
+        self.ri
+            .push(RistrettoPoint::vartime_multiscalar_mul(ri_scalars, ri_points));
```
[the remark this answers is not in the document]
hansieodendaal: As per above

src/range_proof.rs
```diff
@@ -0,0 +1,573 @@
...
+        let h_base_compressed = statement.generators.h_base_compressed();
+        let g_base_compressed = statement.generators.g_base_compressed();
+        let (hi_base, gi_base) = (statement.generators.hi_base(), statement.generators.gi_base());
+
+        transcript.domain_separator(b"Bulletproofs+", b"Range Proof");
+        transcript.validate_and_append_point(b"H", &h_base_compressed)?;
+        transcript.validate_and_append_point(b"G", &g_base_compressed)?;
+        transcript.append_u64(b"N", bit_length as u64);
+        transcript.append_u64(b"M", aggregation_factor as u64);
+        for item in statement.commitments_compressed.clone() {
```
mikethetike, line 135: unnecessary clone here

src/range_proof.rs
```diff
@@ -0,0 +1,573 @@
...
+            transcript.append_point(b"Ci", &item);
+        }
+        for item in statement.minimum_value_promises.clone() {
```
mikethetike, line 138: And here

src/range_proof.rs
hansieodendaal: Thanks for the chat, going to hide the transcript inside the prove method

PR #4, opened 2022-05-05
feat: add serialize/de-serialize

hansieodendaal:
- Added serialize and deserialize implementations
- Expanded the unit tests to test this
- Small refactoring

### DIFF
--- a/src/range_proof.rs
+++ b/src/range_proof.rs
@@ -622,9 +622,9 @@ impl RangeProof {
 
     /// Deserializes the proof from a byte slice
     pub fn from_bytes(slice: &[u8]) -> Result<RangeProof, ProofError> {
-        if slice.len() % 32 != 0 {
+        if slice.len() % 32 != 0 || slice.is_empty() || slice.len() / 32 < 6 {
             return Err(ProofError::InvalidLength(
-                "Serialized proof bytes must be a factor of 32".to_string(),
+                "Invalid serialized proof bytes length".to_string(),
             ));
         }
         let num_elements = slice.len() / 32;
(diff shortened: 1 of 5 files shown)


src/utils/generic.rs
```diff
@@ -21,43 +24,56 @@ pub fn nonce(seed_nonce: &Scalar, label: &str, index: Option<usize>) -> Result<S
     let encoded_label = label.as_bytes();
     if encoded_label.len() > 16 {
         // See https://www.blake2.net/blake2.pdf section 2.8
-        return Err(ProofError::InvalidLength("nonce label".to_string()));
-    };
-    let hasher = if let Some(salt) = index {
+        Err(ProofError::InvalidLength("nonce label".to_string()))
+    } else if let Some(salt) = index {
```
SWvheerden, line 28: I prefer the old code here. 
We introduce more nesting by removing the return

src/utils/generic.rs
```diff
@@ -21,43 +24,56 @@ pub fn nonce(seed_nonce: &Scalar, label: &str, index: Option<usize>) -> Result<S
...
+            encoded_label,
+        )))
+    }
 }
 
 /// Decompose a given value into a vector of scalars for the required bit length
 pub fn bit_vector_of_scalars(value: u64, bit_length: usize) -> Result<Vec<Scalar>, ProofError> {
     if !bit_length.is_power_of_two() || bit_length > RangeProof::MAX_BIT_LENGTH {
-        return Err(ProofError::InvalidLength(
+        Err(ProofError::InvalidLength(
```
SWvheerden, line 52: Same here, nested code reads and follows more difficult than non nested code.

SWvheerden (commented):
Looks good to me. 
I just dont like the additional nested if's, 
They are a bad pattern

src/utils/generic.rs
hansieodendaal: This is the way @DenisKolodin proposed to get rid of additional `return` statements.

src/utils/generic.rs
hansieodendaal: See above

src/utils/generic.rs
hansieodendaal: Reverted

src/utils/generic.rs
hansieodendaal: Reverted

src/range_proof.rs
```diff
@@ -600,6 +601,72 @@ impl RangeProof {
...
+                "Serialized proof bytes must be a factor of 32".to_string(),
+            ));
+        }
+        let num_elements = slice.len() / 32;
+        if (num_elements - 6) % 2 != 0 {
+            return Err(ProofError::InvalidLength(
+                "Serialized proof has incorrect number of elements".to_string(),
+            ));
+        }
+        let n = (num_elements - 6) / 2;
```
sdbondi, line 636: For `slice.len() == 0` this wraps around to a very large number.

sdbondi (commented):
Looks good, I think that this needs more test cases, byte de/serialisation in general should always have extensive tests and ideally fuzzing.

src/range_proof.rs
hansieodendaal: Thanks, fixed and tested

sdbondi (approved):
Looks good!

PR #5, opened 2022-05-11
feat: add extended commitments

hansieodendaal:
Add extended commitments

- Added ~two~ **four** additional degrees on the the Pedersen commitment, called 'extended commitments'.
- Added extended degree Pedersen generators unit tests.
- Refactored overly long functions to improve read-ability.
- Added extended commitments to the unit tests.
- Added extended commitments benchmark tests and documentation.
- **Edit:** Removed pre-calculated basepoints compression in the Pedersen generators in favour of a lazy static implementation - benchmarks showed these to be of equal performance.
- **Edit:** Added three explicit methods for verification: verify and retrieve masks, verify but do not retrieve masks, retrieve masks only. This posed significant gains for public verifying and private mask retrieval alike.
- **Edit:** Added 1x byte to the proof as the extension degree to enable deserialization as too many degrees of freedom were introduced to automatically detect three vector lengths.


src/generators/pedersen_gens.rs
```diff
@@ -23,58 +23,238 @@ use crate::protocols::ristretto_point_protocol::RistrettoPointProtocol;
...
+                .as_bytes()
+        )
+        .compress()
+        .as_bytes()
+    );
+    static ref RISTRETTO_BASEPOINT_COMPRESSED_BLINDING_2: CompressedRistretto =
+        (*RISTRETTO_BASEPOINT_POINT_BLINDING_2).compress();
+    static ref RISTRETTO_BASEPOINT_COMPRESSED_BLINDING_3: CompressedRistretto =
+        (*RISTRETTO_BASEPOINT_POINT_BLINDING_3).compress();
 }
```
AaronFeickert, line 72: Recommend simply hashing a unique indexed string to generate these points, rather than iterated hashing. This is less likely to result in unintended collisions or reuse elsewhere.

src/generators/pedersen_gens.rs
```diff
@@ -21,60 +21,294 @@ use crate::protocols::ristretto_point_protocol::RistrettoPointProtocol;
...
 
 impl PedersenGens {
-    // Pre-calculated '(*RISTRETTO_BASEPOINT_POINT_BLINDING).compress()'
-    const RISTRETTO_BASEPOINT_COMPRESSED_BLINDING: CompressedRistretto = CompressedRistretto([
-        140, 146, 64, 180, 86, 169, 230, 220, 101, 195, 119, 161, 4, 141, 116, 95, 148, 160, 140, 219, 127, 68, 203,
-        205, 123, 70, 243, 64, 72, 135, 17, 52,
-    ]);
-
     /// Creates a Pedersen commitment using the value scalar and a blinding factor.
```
AaronFeickert, line 86: Comment references only a single mask, but a vector of masks is now used.

src/inner_product_round.rs
```diff
@@ -63,81 +65,101 @@ impl<'a> InnerProductRound<'a> {
...
         if gi_base.is_empty() || hi_base.is_empty() || ai.is_empty() || bi.is_empty() || y_powers.is_empty() {
-            Err(ProofError::InvalidLength(
+            return Err(ProofError::InvalidLength(
                 "Vectors gi_base, hi_base, ai, bi and y_powers cannot be empty".to_string(),
-            ))
-        } else if !(hi_base.len() == n && ai.len() == n && bi.len() == n) || (y_powers.len() != (n + 2)) {
-            Err(ProofError::InvalidArgument(
+            ));
+        }
+        if !(hi_base.len() == n && ai.len() == n && bi.len() == n) || (y_powers.len() != (n + 2)) {
```
AaronFeickert, line 84: Doesn't `gi_base` need its length checked as well?

src/range_proof.rs
```diff
@@ -383,7 +418,7 @@ impl RangeProof {
         two_n_minus_one -= Scalar::one();
 
         // Weighted coefficients for common generators
-        let mut g_base_scalar = Scalar::zero();
+        let mut g_base_scalar = vec![Scalar::zero(); extension_degree];
```
AaronFeickert, line 421: Perhaps pluralize this to make it consistent with the other vectors.

src/utils/generic.rs
```diff
@@ -11,28 +11,41 @@ use core::{
...
-        Blake2b::with_params(&seed_nonce.to_bytes(), &encoded_index, encoded_label)
+        Blake2b::with_params(&seed_nonce.to_bytes(), &encoded_salt, encoded_label)
     } else {
         Blake2b::with_params(&seed_nonce.to_bytes(), &[], encoded_label)
     };
+    if let Some(salt) = index_k {
+        let encoded_salt = salt.to_le_bytes();
+        if encoded_salt.len() > 16 {
+            // See https://www.blake2.net/blake2.pdf section 2.8
+            return Err(ProofError::InvalidLength("nonce index_k".to_string()));
```
AaronFeickert, line 45: Since this index is being used as input to the hash function (and not directly as a salt parameter), this length limitation does not apply.

You could also concatenate fixed-length encodings of `j` and `k` and use this as the salt parameter; such an approach seems more canonical and in the spirit of the hash function parameter set, but it does imply further restriction on the index values.

src/utils/generic.rs
```diff
@@ -11,28 +11,41 @@ use core::{
...
+        };
+        hasher.update(&encoded_salt);
+    }
 
     Ok(Scalar::from_hasher_blake2b(hasher))
```
AaronFeickert, line 50: Has it been verified that the hash output with no `k` set (which should hash the empty string) is in fact unique across different nonces? It should be, but it might be nice to test this specifically to ensure the hasher is working as expected when no input is passed to it.

AaronFeickert (changes requested):
See individual comments.

AaronFeickert:
Are there any issues with the minimum value bounds and verifier behavior?

AaronFeickert:
> Are there any issues with the minimum value bounds and verifier behavior?

For example, checking that the minimum value does not exceed the bound from the statement bit length is an easy way to avoid attempted verification of a proof that is guaranteed to fail.

src/inner_product_round.rs
hansieodendaal: Indeed it does, equal to `n` for that statement (`let n = gi_base.len();`).

src/utils/generic.rs
hansieodendaal: Good observation; I removed the length checks on both indexes as they are limited by the length of the byte representation of `usize::MAX`, which is `< 16`, and added a parameter length test. 

Interestingly enough, if the `persona` or `salt` parameters exceed 16 bytes, unhandled exceptions occur, so we have to do the length check ourselves.

Example exception passing 17 bytes into the `persona` parameter:
```
assertion failed: persona.len() <= length
thread 'utils::generic::tests::test_blake2b_parameter_lengths' panicked at 
   'assertion failed: persona.len() <= length', 
   .cargo\registry\src\github.com-1ecc6299db9ec823\blake2-0.9.2\src\blake2b.rs:4:1
```

src/utils/generic.rs
hansieodendaal: Yes, that is specifically tested.

hansieodendaal: Added additional tests

hansieodendaal:
> does not exceed the bound from the statement bit length

No short-circuit verification tests for providing the wrong minimum value. Maybe stating the obvious here, but the statement must supply the correct minimum value. A proof constructed with `value = 10 000` and `min_value = 100` will fail if a statement with `min_value = 101` is supplied.

AaronFeickert:
> > does not exceed the bound from the statement bit length
> 
> No short-circuit verification tests for providing the wrong minimum value. Maybe stating the obvious here, but the statement must supply the correct minimum value. A proof constructed with `value = 10 000` and `min_value = 100` will fail if a statement with `min_value = 101` is supplied.

Right, but if the specified minimum value were to exceed the maximum value permitted by the statement bit length, attempting to continue verification (rather than simply failing immediately) seems unnecessary, since the verifier can tell right away that the verification _must_ fail.

AaronFeickert:
> > does not exceed the bound from the statement bit length
> 
> No short-circuit verification tests for providing the wrong minimum value. Maybe stating the obvious here, but the statement must supply the correct minimum value. A proof constructed with `value = 10 000` and `min_value = 100` will fail if a statement with `min_value = 101` is supplied.

Sure, but there are situations where the verifier can use public statement data to determine a guaranteed failure. If the minimum value exceeds `2^N - 1` (where `N` is the statement bit length), then the proof will always fail; no need to run the rest of the verification logic.

src/range_proof.rs
```diff
@@ -613,10 +643,13 @@ impl RangeProof {
                 points.push(ri[j]);
             }
         }
+        if extract_masks == ExtractMasks::ONLY {
+            return Ok(masks);
+        }
```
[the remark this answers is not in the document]
hansieodendaal: Yip, good point, easily enough to have a separate user interface to perform mask recovery only.

src/utils/generic.rs
```diff
@@ -11,41 +11,39 @@ use core::{
...
+    let mut key = Vec::with_capacity(51); // 1 + 32 + 1 + 8 + 1 + 8
+    key.push(0u8); // Add one byte to initialize the vector
+    key.append(&mut seed_nonce.to_bytes().to_vec());
+    if let Some(index) = index_j {
+        key.append(&mut b"j".to_vec()); // Domain separated index label
+        key.append(&mut index.to_le_bytes().to_vec());
+    }
+    if let Some(index) = index_k {
+        key.append(&mut b"k".to_vec()); // Domain separated index label
+        key.append(&mut index.to_le_bytes().to_vec());
```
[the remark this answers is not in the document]
hansieodendaal: So `seed_nonce.to_bytes()` is a fixed length encoding of input `seed_nonce: &Scalar`, so a byte representation of `seed_nonce = some_arbitrary_seed` can never yield the concatenated `some_arbitrary_seedj0k0`. I also introduced  a unit test for collisions of `seed_nonce = some_arbitrary_seed, j = Some(value1), k = None` and `seed_nonce = some_arbitrary_seed, j = None, k = Some(value1)`.

That being said, your DM about fixed-length encodings of inputs `index_j: Option<usize>` and `index_k: Option<usize>` or proper use of domain separation labels is definitely missing.

I would prefer using fixed length concatenation with proper domain separation labels as the hash input as I saw a noticeable decrease in performance when I concatenated both indexes as the salt (although there might be a reasonable explanation for this).

hansieodendaal:
> Right, but if the specified minimum value were to exceed the maximum value permitted by the statement bit length, attempting to continue verification (rather than simply failing immediately) seems unnecessary, since the verifier can tell right away that the verification _must_ fail.

In the tari blockchain implementation using any bit length other than `N = 64` is of academic value only as we have to cater for values being `u64`, however, from a library point of view introducing the sanity check you propose is certainly worth while so I will add it.

AaronFeickert:
> > Right, but if the specified minimum value were to exceed the maximum value permitted by the statement bit length, attempting to continue verification (rather than simply failing immediately) seems unnecessary, since the verifier can tell right away that the verification _must_ fail.
> 
> In the tari blockchain implementation using any bit length other than `N = 64` is of academic value only as we have to cater for values being `u64`, however, from a library point of view introducing the sanity check you propose is certainly worth while so I will add it.

Yeah. Plus, the library is written with variable bit length support, so it makes sense IMO to properly support this.

src/range_proof.rs
```diff
@@ -16,13 +16,26 @@ use serde::{de::Visitor, Deserialize, Deserializer, Serialize, Serializer};
...
-    utils::generic::{bit_vector_of_scalars, nonce, read32},
+    utils::generic::{bit_vector_of_scalars, nonce, read32, read8},
+    PedersenGens,
 };
 
+/// Optionally extract masks when verifying the proofs
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum ExtractMasks {
+    /// No masks will be recovered (e.g. as a public entity)
+    NO,
```
mikethetike, line 32: I would have thought Clippy would complain about this naming. Perhaps we're missing a rule in lints

mikethetike (commented):
I'll differ to @AaronFeickert's opinion on correctness. I only have a few minor comments.

I think a next step would be to refactor the code to take in a generic curve instead of RistrettoPoint and the PedersonGens. This will allow a library like `tari_crypto` to use it's own generators

src/range_proof.rs
hansieodendaal: I have added `'clippy::upper_case_acronyms',`, but to no avail, not picked up.

AaronFeickert:
Any flamegraph-type results on batch verification? Would be nice to see if there any particular spots that should be the focus of optimizations.

hansieodendaal:
I have left flamegraph optimizations for a next PR...

mikethetike (approved):
Happy for now.  As I said, a follow up PR should make it more generic for consumers of this library to use their own generators and curves

PR #6, opened 2022-05-19
feat: make BP+ agnostic to the curve point

sdbondi:
This PR refactors BP+ to be agnostic to the curve used. Implementers are required to implement `CurvePointProtocol` and provide commitment base points.

I was not able to implement RangeProofs for Edwards because Elligator is not currently implemented for that curve group (from_uniform_bytes cannot be implemented unless this is in place)


src/range_proof.rs
```diff
@@ -83,7 +98,14 @@ pub struct RangeProof {
...
 ///     let extension_degree = ExtensionDegree::Zero;
-///     let generators = RangeParameters::init(bit_length, aggregation_size, extension_degree).unwrap();
+///     let pc_gens = ristretto::create_pedersen_gens_with_extension_degree(extension_degree);
+///     let generators = RangeParameters::init(bit_length, aggregation_size, pc_gens).unwrap();
+///     let generators = RangeParameters::init(
+///         bit_length,
+///         aggregation_size,
+///         ristretto::create_degree_zero_pedersen_gens(),
+///     )
+///     .unwrap();
```
hansieodendaal, line 108:
```suggestion
```
This is a duplicate line - the former has clearer intent as `extension_degree` is re-used further down.

src/ristretto.rs
```diff
@@ -0,0 +1,257 @@
...
+    fn compress(&self) -> Self::Compressed {
+        RistrettoPoint::compress(self)
+    }
+}
+
+/// Generate `PedersenGens` with an `ExtensionDegree` of zero
+pub fn create_degree_zero_pedersen_gens() -> PedersenGens<RistrettoPoint> {
+    create_pedersen_gens_with_extension_degree(ExtensionDegree::Zero)
+}
+
```
hansieodendaal, line 62:
```suggestion
```

I think we can away with this interface as the intent is quite clear using the default function call.

src/ristretto.rs
```diff
@@ -0,0 +1,257 @@
...
+        }
+    }
+
+    #[test]
+    fn test_create_degree_zero_pederson_gens() {
+        assert_eq!(
+            create_pedersen_gens_with_extension_degree(ExtensionDegree::Zero),
+            create_degree_zero_pedersen_gens()
+        );
+    }
```
hansieodendaal, line 224:
```suggestion
```
As per above

hansieodendaal (commented):
LGTM, just a few comments. I believe this will make implementation in `tari-crypto` a lot easier.

src/ristretto.rs
sdbondi: Would you rather rename this to `default_pedersen_gens`? Happy to do so.

The `Default` impl won't work on PedersenGens<P> because there is no generic idea of "default" for base points. If the idea of default PedersonGens were integral to bp+ we could create a new trait for this. I think there could well be further iterations on how we do generators.

hansieodendaal: I actually meant having `create_pedersen_gens_with_extension_degree` is clear enough, or maybe as you suggest `default_pedersen_gens`, either will do, but my first choice would be just to remove that helper function.

sdbondi: ah ok cool, yeah completely agree

src/generators/generators_chain.rs
```diff
@@ -4,17 +4,21 @@
...
 use sha3::{Sha3XofReader, Shake256};
 
+use crate::traits::FromUniformBytes;
+
 /// The `GeneratorsChain` creates an arbitrary-long sequence of orthogonal generators.  The sequence can be
 /// deterministically produced starting with an arbitrary point.
-pub struct GeneratorsChain {
+pub struct GeneratorsChain<P> {
     reader: Sha3XofReader,
+    _phantom: PhantomData<P>,
```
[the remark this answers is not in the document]
sdbondi: :ghost:

hansieodendaal (approved):
LGTM

PR #8, opened 2022-05-24
chore: reduce benches time

hansieodendaal:
Reduced benches time


hansieodendaal:
@mikethetike something other than time-out wrong with the benchmarks - it now completes all expected test successfully in ~12.5 minutes with the last test being `batched_range_proof_verification/Batched 64-bit BP+ verify 8 deg Zero masks VerifyOnly`, but it still exits with an error. I expect something is off with `benchmark.yml`

PR #10, opened 2022-05-30
feat: relax commit bounds

hansieodendaal:
Relaxed commit bounds by allowing generators with more extension degrees (that remain unused) as is strictly necessary for the resulting multi-exponentiation multiplication.


SWvheerden (approved):
LGTM

stringhandler (approved):

Issue #9, opened 2022-05-30
Update `commit` to accept references


hansieodendaal:
It's unfortunate that the BP+ API forces copies :(

There are a few ways to remove this need:

Update `commit` in the BP+ API to accept `I: Iterator<Item=&Scalar>`, then you can pass in `k_i.iter().map(|k| &k.0)`

Or take something like 
``` rust
pub fn commit<S: AsRef<Scalar>>(&self, value: Scalar, blindings: &[S]) -> Result<P, ProofError> 
```

Either way, it requires changes to the BP+ repo, which is not something we can do in this PR.

_Originally posted by @CjS77 in https://github.com/tari-project/tari-crypto/pull/99#discussion_r883695065_

hansieodendaal:
See #12

PR #12, opened 2022-05-31
fix: avoids copying in the commit method

DenisKolodin:
The PR improves `PedersenGens::commit` method:

- [x] Internal copying: avoid copying values inside the method (use iters)
- [ ] External copying: avoid copying on calls (support `AsRef<Scalar>` args)


hansieodendaal (commented):
LGTM - solves #9 
(tests need to be updated, `.commit(` in tests needs a reference, as per `cargo clippy --all-targets -- -D warnings`)

PR #14, opened 2022-06-05
feat: add methods to retrieve extension degree

hansieodendaal:
Add methods to retrieve extension degree from a proof and a serialized proof. This will assist verifiers in ensuring that the statement contains Pedersen generators of the correct extension degree before attempting verification.

### DIFF
--- a/src/range_proof.rs
+++ b/src/range_proof.rs
@@ -33,7 +33,7 @@ use crate::{
     range_witness::RangeWitness,
     traits::{Compressable, Decompressable, FixedBytesRepr},
     transcripts,
-    utils::generic::{bit_vector_of_scalars, nonce, read32, read8},
+    utils::generic::{bit_vector_of_scalars, nonce, read_1_byte, read_32_bytes},
 };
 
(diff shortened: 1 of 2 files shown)


src/range_proof.rs
```diff
@@ -890,6 +895,16 @@ where
...
     }
+
+    /// Helper function to return the serialized proof's extension degree
+    pub fn extension_degree_from_proof_bytes(slice: &[u8]) -> Result<ExtensionDegree, ProofError> {
+        if slice.is_empty() || (slice.len() - 1) % 32 != 0 {
+            return Err(ProofError::InvalidLength(
+                "Invalid serialized proof bytes length".to_string(),
+            ));
+        }
+        ExtensionDegree::try_from(read8(&slice[0..])[0] as usize)
```
stringhandler, line 906: This surely only needs to be a byte

hansieodendaal: Yip, actually the name is confusing. Fixed the name.

stringhandler (approved):

PR #17, opened 2022-09-29
fix: remove creation of txt file

SWvheerden:
Remove the creation of a text file containing benchmarks results


CjS77:
The change appears to break the benchmarks though?

CjS77:
I think there are bigger issues than the txt file creation. I've disabled the BMs for now until they can be more broadly resolved

Issue #18, opened 2022-10-03
Consider precomputation tables for verification efficiency


AaronFeickert:
The `dalek` curve library multiscalar multiplication operations support the use of precomputation tables. When used, they offer a significant efficiency benefit. The downside is the memory occupied by the tables (according to the library documentation, around 30 kB per table) and the generation time. Since the inner product argument generators are global, their tables could be computed just once.

AaronFeickert:
Unfortunately, the `dalek` library does not support partial precomputation, where you may only wish to use a subset of static points for a particular computation. This would be highly beneficial for this use case, since you could precompute tables for fixed generators for all supported proof sizes.

Fortunately, it's trivial to modify the `dalek` library to support this (the change is two lines, plus tests). The API also does not change, since the effect of mismatched static point and scalar iterators is currently a panic (and a documentation comment warning against doing this).

AaronFeickert:
Some [very preliminary work](https://github.com/AaronFeickert/bulletproofs-plus/tree/precomputation) towards this.

PR #19, opened 2022-10-11
feat: use precomputation on (most) fixed generators

AaronFeickert:
Uses precomputation of fixed generator vectors to speed up verification, at the cost of storing precomputation tables between verification operations. This doesn't use precomputation on the Pedersen generators, since those can be set independently of the others, and we can't mix-and-match precomputation tables due to upstream limitations.

Note that this requires and uses a custom curve library fork. The fork supports partial precomputation by removing an existing restriction about matching the number of static points and scalars used for precomputation evaluation. It also implements `Clone` on the underlying types used for precomputation. This is unfortunate, since due to their size (several megabytes in total) such tables should almost certainly never be cloned. However, it's done for the reason explained below.

The generator tables are wrapped in an `Arc` for shared ownership. This is done because precomputation evaluation is an instance method on a precomputation type, not a static method that takes a reference to the underlying tables. I have no idea why this design was chosen (static methods are used for other types of multiscalar multiplication), especially because there's no mutation involved. But because of this, the verifier needs to own the precomputation structure containing the tables, even though those tables are expected to be reused (that's the entire point of precomputation). Using an `Arc` takes care of this nicely, and avoids cloning.

However, apparently `#[derive(Clone)]` only plays nicely with structs if all included generic types implement `Clone`, which means even though cloning the table `Arc` isn't any kind of deep copy, we can't use that attribute unless the precomputation tables implement `Clone`. Manually implementing `Clone` on the containing struct is a headache, so it seemed easier just to add `#[derive(Clone)]` at the curve library level.

This means it's probably _very important_ to ensure that precomputation tables are used very carefully to avoid unintended cloning. I did some testing and confirmed that the current implementation handles this as expected, and won't clone any of the tables, despite the compiler requiring they implement `Clone`.

Closes [issue #18](https://github.com/tari-project/bulletproofs-plus/issues/18).

### DIFF
--- a/src/range_parameters.rs
+++ b/src/range_parameters.rs
@@ -3,7 +3,10 @@
 
 //! Bulletproofs+ range parameters (generators and base points) needed for a batch of range proofs
 
-use std::fmt::{Debug, Formatter};
+use std::{
+    fmt::{Debug, Formatter},
+    sync::Arc,
+};
 
 use crate::{
     errors::ProofError,
@@ -12,12 +15,12 @@ use crate::{
         pedersen_gens::{ExtensionDegree, PedersenGens},
     },
     range_proof::MAX_RANGE_PROOF_BIT_LENGTH,
-    traits::{Compressable, FromUniformBytes},
+    traits::{Compressable, FromUniformBytes, Precomputable},
 };
 
 /// Contains all the generators and base points needed for a batch of range proofs
 #[derive(Clone)]
-pub struct RangeParameters<P: Compressable> {
+pub struct RangeParameters<P: Compressable + Precomputable> {
     /// Generators needed for aggregating up to `m` range proofs of up to `n` bits each.
     bp_gens: BulletproofGens<P>,
     /// The pair of base points for Pedersen commitments.
@@ -25,7 +28,7 @@ pub struct RangeParameters<P: Compressable> {
 }
 
 impl<P> RangeParameters<P>
-where P: FromUniformBytes + Compressable + Clone
+where P: FromUniformBytes + Compressable + Clone + Precomputable
 {
     /// Initialize a new 'RangeParameters' with sanity checks
     pub fn init(bit_length: usize, aggregation_factor: usize, pc_gens: PedersenGens<P>) -> Result<Self, ProofError> {
@@ -107,11 +110,6 @@ where P: FromUniformBytes + Compressable + Clone
         self.hi_base_iter().collect()
     }
 
-    /// Return the non-public value bulletproof generator references
-    pub fn hi_base_copied(&self) -> Vec<P> {
-        self.hi_base_iter().cloned().collect()
-    }
-
     /// Return the non-public mask iterator to the bulletproof generators
     pub fn gi_base_iter(&self) -> impl Iterator<Item = &P> {
         self.bp_gens.g_iter(self.bit_length(), self.aggregation_factor())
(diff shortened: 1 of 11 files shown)


stringhandler:
I think you will struggle to get that PR into dalek, but worth a try. Looks ok at first glance otherwise

AaronFeickert:
> I think you will struggle to get that PR into dalek, but worth a try. Looks ok at first glance otherwise

Agreed that it seems very unlikely for an upstream PR to be accepted, and I didn't plan on doing one. I think it would need to remain a fork. However, keeping a maintained fork would also allow for more flexibility on future changes to the curve library.

AaronFeickert:
This will be ready to review once a [curve library PR](https://github.com/tari-project/curve25519-dalek/pull/1) is merged and tagged, at which point this PR can use it as a dependency.

stringhandler:
@AaronFeickert I've merged the curve library PR and created a tag

src/range_parameters.rs
```diff
@@ -109,7 +112,7 @@ where P: FromUniformBytes + Compressable + Clone
 
     /// Return the non-public value bulletproof generator references
     pub fn hi_base_copied(&self) -> Vec<P> {
-        self.hi_base_iter().cloned().collect()
+        self.hi_base_iter().copied().collect()
```
SWvheerden, line 115: I made an issue about the memory cloning in BP+, this is one of them, is there a reason we copy it here and not just return a ref to it?
If the callee really wants to clone it, they can. Here it is forced upon the callee.

src/range_proof.rs
```diff
@@ -186,7 +187,7 @@ impl<P> RangeProof<P>
 where
     for<'p> &'p P: Mul<Scalar, Output = P>,
     for<'p> &'p P: Add<Output = P>,
-    P: CurvePointProtocol,
+    P: Copy + CurvePointProtocol + Precomputable,
```
SWvheerden, line 190: personally preference, I don't like `Copy` here
The scalars are massive and this can lead to bad design choices. By forcing the developer to call `clone`  each time a clone/copy is desired it becomes a conscious choice on each clone/copy. 

looking at the new code where the clones/copy happens, it now hides the fact that the code is doing an expensive operation.

src/range_statement.rs
```diff
@@ -29,7 +29,7 @@ pub struct RangeStatement<P: Compressable> {
     pub seed_nonce: Option<Scalar>,
 }
 
-impl<P: Compressable + FromUniformBytes + Clone> RangeStatement<P> {
+impl<P: Copy + Compressable + FromUniformBytes + Clone + Precomputable> RangeStatement<P> {
```
SWvheerden, line 32: Same reason as above, dont like the copy here

SWvheerden (commented):
This looks good. 
I am mostly happy with it except for the copy. Expensive types should not have copy as an implementation. 

The reason is two-fold:
- Types that impl copy are not moved when used in a function, they are copied
- If a type impl copy it hides expensive copying by always doing mem copy and not doing a move which is more desirable and or pointing out to the dev that a clone is happening here,

src/range_parameters.rs
AaronFeickert: Removed in a new update.

src/range_proof.rs
AaronFeickert: Removed in a new update.

src/range_statement.rs
AaronFeickert: Removed in a new update.

SWvheerden (approved):
This is more going in the right direction

AaronFeickert:
This is now an [upstream PR](https://github.com/dalek-cryptography/curve25519-dalek/pull/546).

PR #23, opened 2022-12-14
chore: update toolchain

CjS77:
A recent serde upgrade broke this crate, so navigating dependency hell
to make everything play nce again. In summary:

- change toolchain to stable
- update workflows
Some workflows need to run on nightly, but update to stable where
possible
- re-introduce --all-features to test workflow, since simd has since
been updated
- add asref and asmut to NonDebug to make clippy shut up


CjS77:
ACK

CjS77:
ACK

PR #26, opened 2023-04-03
fix: fix rand_core inequality

hansieodendaal:
`merlin = "3"` uses `rand_core = "6"` instead of `rand_core = "5"` like the rest of the code base


hansieodendaal:
Awesome thanks

Issue #27, opened 2023-04-05
Computational offloading


AaronFeickert:
This issue informally describes a method for allowing a limited device to offload expensive Bulletproof+ (BP+) range proof computation to another semi-trusted device. It addresses the important use case where a computationally-limited device (the _signer_) knows the mask for a commitment for which it wishes to produce a range proof, but is unable to produce the complete proof itself because of expensive operations. The signer has an associated device (the _helper_) that is not computationally limited, but which should not learn the mask. The signer wishes to collaborate with the helper to produce the desired range proof.

For this analysis, we assume that the signer is trusted, and that the helper is honest but curious. This means the helper will always follow a given protocol, but can use information at its disposal in an attempt to recover secret data.

The notation we use is from the [BP+ preprint](https://eprint.iacr.org/2020/735), specifically from Figure 2 on page 16.

## Design

The signer and helper want to produce a BP+ range proof for a Pedersen commitment $V = vG + \gamma H$. The signer and helper both know $V$ and the value $v$, both only the signer knows the mask $\gamma$.

The signer samples a scalar $\alpha$ uniformly at random and sends $\alpha H$ to the helper. The helper uses this as the masking component to the bit commitment $A$ it computes. It sends $A$ to the signer.

The signer uses $V$ and $A$ to compute the Fiat-Shamir challenge $y$, and then computes $\widehat{\alpha} = \alpha + \gamma y^{n+1}$. It sends $\widehat{\alpha}$ to the helper.

The helper uses $\widehat{\alpha}$ to complete the range proof, which is a standard BP+ range proof.

## Mask extraction

The current BP+ range proof implementation supports mask extraction, where the prover and verifier share a seed nonce that is used to produce certain proof elements. Knowledge of this value is sufficient for the verifier to extract the commitment mask from the proof.

The technique cannot be directly applied to this design. If the signer uses the same seed nonce to derive $\alpha$ as the helper uses to derive other proof elements, the helper can use this value to trivially recover the mask.

A modification suitable for this design requires the use of two seed nonces. One of these values is known to the signer, and is used to derive $\alpha$. The other value is known to the helper, and is used to derive all other proof elements as required. This ensures that the helper cannot learn $\alpha$, and therefore cannot recover the mask.

When the verifier wishes to extract the mask, it must use both seed nonces.

## Security

[...]

PR #29, opened 2023-04-20
feat!: proof offloading

AaronFeickert:
A [recent issue](https://github.com/tari-project/bulletproofs-plus/issues/27) suggests a design for proof offloading. This PR implements it using a state machine design with a wrapper for existing proving functionality. Closes [issue 27](https://github.com/tari-project/bulletproofs-plus/issues/27).

The linked design assumes that a signer (who knows commitment masks) and a helper (who only knows the commitment values and commitments) want to collaborate to produce a range proof. We implement their interaction using an informal state machine, where each state transition consumes input message data and a previous state, and generates output message data and a new state. This means that both the signer and helper need the capability to retain this state during the proof generation process.

Because the signer does not want the helper to be able to obtain the mask, we add an additional seed nonce. The signer uses one seed nonce during its operations, and the helper uses the other. During verification, a designated verifier who knows both seeds can extract the mask. To help make this work more cleanly, we move the seed nonces to their own separate `RangeSeedNonce`, which is no longer part of `RangeStatement`. This also seems more in line with the functioning of seed nonces, which aren't strictly part of the range statement for non-designated verifiers. We also change each of the seed nonces to be a byte vector, which is more generic.

We keep the existing proving functionality, where the signer and helper are the same entity. This is accomplished by the use of a simple wrapper that executes the state machine transitions for the signer and helper itself, and keeps the code cleaner.

Note that the linked design has security implications relating to rewinding attacks. The state machine design is intended to help mitigate these, but any nontrivial implementations should carefully consider the implications of such attacks.

BREAKING CHANGE: Changes the prover and verifier APIs. Updates how seed nonces are used in mask recovery. Existing range proofs using seed nonces will fail to recover masks.

### DIFF
--- a/src/inner_product_round.rs
+++ b/src/inner_product_round.rs
@@ -52,7 +52,7 @@ pub struct InnerProductRound<'a, P> {
 
     // Seed for mask recovery
     round: usize,
-    seed_nonce: Option<Scalar>,
+    seed_nonce_helper: Option<Vec<u8>>,
 }
 
 impl<'a, P: 'a> InnerProductRound<'a, P>
@@ -74,7 +74,7 @@ where
         alpha: Vec<Scalar>,
         y_powers: Vec<Scalar>,
         transcript: &'a mut Transcript,
-        seed_nonce: Option<Scalar>,
+        seed_nonce_helper: Option<Vec<u8>>,
         aggregation_factor: usize,
     ) -> Result<Self, ProofError> {
         let n = gi_base.len();
@@ -112,7 +112,7 @@ where
             ri: Vec::with_capacity(n * aggregation_factor + 2),
             transcript: transcript.into(),
             round: 0,
-            seed_nonce,
+            seed_nonce_helper,
         })
     }
 
@@ -130,10 +130,10 @@ where
                 Vec::with_capacity(extension_degree),
                 Vec::with_capacity(extension_degree),
             );
-            if let Some(seed_nonce) = self.seed_nonce {
+            if let Some(seed_nonce_helper) = &self.seed_nonce_helper {
                 for k in 0..extension_degree {
-                    d.push((nonce(&seed_nonce, "d", None, Some(k)))?);
-                    eta.push((nonce(&seed_nonce, "eta", None, Some(k)))?);
+                    d.push((nonce(seed_nonce_helper, "d", None, Some(k)))?);
+                    eta.push((nonce(seed_nonce_helper, "eta", None, Some(k)))?);
                 }
             } else {
                 for _k in 0..extension_degree {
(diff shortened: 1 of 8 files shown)


AaronFeickert:
I'm still in the process of testing this across the broader codebase, especially relating to the use of two seed nonces in a safe way.

src/range_proof.rs
```diff
@@ -194,93 +252,203 @@ where
...
+                a_ri,
+                a,
+            },
+        ))
+    }
+
+    /// The signer receives the bit commitment message from the helper and produces the masked masks (yes, you read that
+    /// correctly)
+    #[allow(clippy::needless_pass_by_value)] // this ensures we consume the helper message
+    fn signer_mask_masks(
```
SWvheerden, line 396: As far as I read the comments and have been following the code, this should run on the signer device?
If that's the case this wont work at all. All of the data inside of the state is too large to run on a ledger. 
We have +- 26kb to work with, h_base + g_base is almost that alone. 

The signer can only do like a simple single math commitment, hashes etc. Transcripts I doubt it can.

src/range_proof.rs
```diff
@@ -351,15 +519,71 @@ where
...
+
+        // Extract the seed nonces for separation
+        let seed_nonce_helper = seed_nonce_pair.as_ref().map(|p| p.seed_nonce_helper.clone());
+        let seed_nonce_signer = seed_nonce_pair.as_ref().map(|p| p.seed_nonce_signer.clone());
+
+        // Extract values from the witness for use by the helper
+        let values = witness.openings.iter().map(|o| o.v).collect::<Vec<u64>>();
+
+        // Prepare the signer bootstrap data
+        let signer_bootstrap = SignerBootstrapData {
```
SWvheerden, line 551: Yeah, this will be too large for a ledger.

SWvheerden (commented):
I have some questions, but we are def on the right track I think

src/inner_product_round.rs
```diff
@@ -52,7 +52,7 @@ pub struct InnerProductRound<'a, P> {
 
     // Seed for mask recovery
     round: usize,
-    seed_nonce: Option<Scalar>,
+    seed_nonce_helper: Option<Vec<u8>>,
```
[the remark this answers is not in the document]
AaronFeickert: I switched from requiring the seed nonce be a `Scalar` to a more general `Vec<u8>` since there never was a cryptographic reason for the former, except to enforce some level of effective entropy.

src/range_proof.rs
AaronFeickert: See other comment about generators.

AaronFeickert:
This design is no longer needed, since the changes described in a [draft RFC](https://github.com/tari-project/rfcs/pull/98) mean that the helper can safely produce the entire range proof itself. Closing.

PR #34, opened 2023-06-19
chore: inner-product protocol cleanup

### LINKED ISSUE #33: Clean up inner-product protocol memory use
The [inner-product protocol](https://github.com/tari-project/bulletproofs-plus/blob/cd7588ee8eaebe862fe9cf5d7c3fd92981703e87/src/inner_product_round.rs#L120-L271) gets a bit messy in its [computation](https://github.com/tari-project/bulletproofs-plus/blob/cd7588ee8eaebe862fe9cf5d7c3fd92981703e87/src/inner_product_round.rs#L212-L238) of `L` and `R` used in each round. In particular, it performs a lot of unnecessary cloning of both scalar and point vectors while preparing multiscalar multiplication inputs.

We can clean this up significantly to reduce memory use and increase efficiency. The variable-time multiscalar multiplication evaluation function can accept reference iterators, so we should use them whenever possible. It's already straightforward to do this for the point vectors.

The scalar vectors are a little trickier. While most values can be referenced directly, there are multiplicative offsets applied to [`a1`](https://github.com/tari-project/bulletproofs-plus/blob/cd7588ee8eaebe862fe9cf5d7c3fd92981703e87/src/inner_product_round.rs#L228) and [`a2`](https://github.com/tari-project/bulletproofs-plus/blob/cd7588ee8eaebe862fe9cf5d7c3fd92981703e87/src/inner_product_round.rs#LL232C19-L232C19) that can't be done without a copy or clone. It may be possible to be clever about these offsets, since a similar offset is [applied later](https://github.com/tari-project/bulletproofs-plus/blob/cd7588ee8eaebe862fe9cf5d7c3fd92981703e87/src/inner_product_round.rs#L256-L259). Even if this wouldn't apply, doing the offsets via a copy or clone is still probably a net win.

AaronFeickert:
The inner-product protocol implementation includes many unnecessary clones when computing per-round multiscalar multiplication evaluations. These are unnecessary, since the evaluation function can accept reference iterators. Further, some multiplicative scalar offsets were being computed twice.

This PR cleans up these parts of the protocol for memory and computational efficiency, The evaluations now use only references, eliminating the clones. It's also more clever about scalar offsets to reduce duplicated computations.

Closes #33.

### DIFF
--- a/src/inner_product_round.rs
+++ b/src/inner_product_round.rs
@@ -183,6 +183,8 @@ where
         } else {
             self.y_powers[n].invert()
         };
+        let a1_offset = a1.iter().map(|s| s * y_n_inverse).collect::<Vec<Scalar>>();
+        let a2_offset = a2.iter().map(|s| s * self.y_powers[n]).collect::<Vec<Scalar>>();
 
         let (mut d_l, mut d_r) = (
             Vec::with_capacity(extension_degree),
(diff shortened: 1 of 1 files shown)


CjS77 (approved):
Oddly, the benchmarks were not faster with this PR on my Macbook Pro (2018). Though it was difficult to replicate due the CPU throttling issues, it looks like the clones are very fast compared to other operations. Maybe the `iter`s have some overhead vs a contiguous block of memory that a vec has 🤷 

Nevertheless, I like this.. it's cleaner and clearly

AaronFeickert:
Yeah, I wondered about whether the use of iterators incurred any overhead, but didn't find anything definitive about this. The scalar computation duplication overhead should be small compared to group operations, and it seems reasonable that it could fall within the noise threshold of benchmarking.

PR #35, opened 2023-06-19
feat: simplify bit vector commitment

AaronFeickert:
Currently, bit vector commitments are computed using [multiscalar multiplication](https://github.com/tari-project/bulletproofs-plus/blob/cd7588ee8eaebe862fe9cf5d7c3fd92981703e87/src/range_proof.rs#L265-L273) against the inner-product generators. However, this can be simplified by using the shiny new [generator precomputation](https://github.com/tari-project/bulletproofs-plus/pull/19) feature.

### DIFF
--- a/src/range_proof.rs
+++ b/src/range_proof.rs
@@ -262,14 +262,10 @@ where
                 Scalar::random_not_zero(rng)
             });
         }
-        let a = P::vartime_multiscalar_mul(
-            alpha.iter().chain(a_li.iter()).chain(a_ri.iter()),
-            statement
-                .generators
-                .g_bases()
-                .iter()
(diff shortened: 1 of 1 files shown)


AaronFeickert:
I saw inconsistent benchmark results with this, despite doing more detailed analysis that compared multiscalar multiplication to the sum-and-difference approach. Comments welcome on this.

AaronFeickert:
Another technique would be to use the new [generator precomputation](https://github.com/tari-project/bulletproofs-plus/pull/19), but initial tests suggested this would still be slower.

CjS77 (approved):
LGTM

PR #38, opened 2023-07-06
chore: update curve library and dependencies

AaronFeickert:
Updates the curve library and associated dependencies.

One change of note is that precomputation tables are now maintained in an `Rc` instead of an `Arc` due to trait bounds.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,7 +45,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: doc
-          args: --features u64_backend
+          args: --all-features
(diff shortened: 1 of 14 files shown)


AaronFeickert:
Adds a couple of commits by @SWvheerden for additional dependency updates.

AaronFeickert:
Since the `digest` functionality was updated, it should be manually checked that hash-based nonces and curve group generators are unchanged from before the update. Any difference would constitute a breaking change.

SWvheerden (approved):

AaronFeickert:
Should a new version be tagged for use in other repositories?

PR #45, opened 2023-07-20
fix: reduce vector capacity

AaronFeickert:
When computing the `d` vector, the prover and verify each allocate too much space for it. This is unnecessary, since the number of scalar elements in the vector is known.

This PR fixes the over-allocation by correcting the requested capacity. This can be checked manually by examining the number of elements pushed to the vector in both the prover and verifier cases.

### DIFF
--- a/src/range_proof.rs
+++ b/src/range_proof.rs
@@ -280,7 +280,7 @@ where
         }
 
         // Compute d efficiently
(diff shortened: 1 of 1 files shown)


CjS77 (approved):
LGTM

PR #47, opened 2023-07-20
test: improve coverage

AaronFeickert:
[Coverage](https://coveralls.io/github/tari-project/bulletproofs-plus?branch=main) for the repository is pretty good, but can be improved by adding tests.

This PR adds tests across the board.

There are a few notable changes made for more robust coverage that should receive extra attention:
- The looping strategy for `Scalar::random_not_zero` has been modified from a `loop` to a `while`. This imposes an extra constant-time equality check, but this is done directly on byte arrays and is very fast.
- The `TranscriptProtocol` trait function `append_scalar` was unused, so it has been removed entirely.
- The `RangeParameter` getter function `bp_gens` was unused, so it has been removed entirely.

PR #48, opened 2023-07-20
fix: promise check in range statement

AaronFeickert:
Range statement initialization requires the number of minimum value promises to match the number of commitments. There's currently a logic error that renders this check incorrect.

This PR fixes the logic. A test will be included in a [broader test PR](https://github.com/tari-project/bulletproofs-plus/pull/47), but the logic change is here for scope.


CjS77 (approved):
Nice

PR #49, opened 2023-07-20
ci: speed up tests

AaronFeickert:
Tests run slowly on CI. This happens for three reasons:
- Integration tests run over a matrix of extension degrees, bit lengths, and batch structures.
- The document test runs over a batch of four proofs.
- CI is configured to run debug and release tests with and without features, even though this crate no longer uses features; this means all `nightly` tests are run identically twice.

This PR speeds up testing and improves CI. It does a few things:
- Reduces the integration test matrix complexity by using smaller batches and fewer bit lengths.
- Simplifies the batching used for the document test.
- Removes duplicate tests from CI.
- Adds `check` and reorders testing jobs for faster failure.
- Removes an unused utility file; this wasn't directly related to testing, but it seemed like a good time to remove it.

These changes reduce CI testing time by about half.


.github/workflows/test.yml
```diff
@@ -20,34 +20,25 @@ jobs:
...
           command: test
-          args: --all-features
-      - name: test/release all features
-        if: ${{ matrix.rust != 'stable' }}
+      - name: test/release
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --release --all-features
+          args: --release
```
CjS77, line 40: You should keep `--all-features` in; it's more future proof

.github/workflows/test.yml
```diff
@@ -20,34 +20,25 @@ jobs:
...
           override: true
-      - name: test/debug
+      - name: check/benches
         uses: actions-rs/cargo@v1
         with:
-          command: test
-      - name: test/release
+          command: check
+          args: --benches
+      - name: check
```
CjS77, line 28: You could  optimise even more by having a single `check--all-features --all-targets` rather than compile for `check` twice. Then you can remove the check/benches job

CjS77 (commented):
Looks fine in general. I want a couple of changes to the CI flow though.

.github/workflows/test.yml
AaronFeickert: I suppose. My thought was that the addition of later features would mean the default feature set wouldn't be separately tested, but agree that this would ensure more complete testing.

.github/workflows/test.yml
AaronFeickert: Nice! I didn't know those flags would cover benches.

CjS77 (approved):
LGTM

PR #50, opened 2023-07-31
chore: manually clone generators

AaronFeickert:
Changes the cloning strategy for `BulletproofGens` from derived to manual. This removes the need for the curve library precomputation struct to support cloning.

### DIFF
--- a/src/generators/bulletproof_gens.rs
+++ b/src/generators/bulletproof_gens.rs
@@ -50,9 +50,13 @@ pub struct BulletproofGens<P: Precomputable> {
     pub(crate) precomp: Arc<P::Precomputation>,
 }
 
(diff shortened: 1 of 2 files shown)


CjS77:
ACK from me. The comment on L#53 is the key. Without that, this would seem superflous.

AaronFeickert:
If the pending upstream curve library PR (which adds partial precomputation) is not accepted or is accepted with cloning, then this PR won't be needed. I'm leaving it as draft for now in case upstream wants to remove cloning, at which point the curve library fork can be maintained more easily.

hansieodendaal (approved):
utACK

SWvheerden (approved):

PR #56, opened 2023-08-07
feat!: add vector length check and improve slicing

### LINKED ISSUE #55: Consider splitting inner-product vectors idiomatically
Currently, the inner-product prover splits vectors using [pairs of slices](https://github.com/tari-project/bulletproofs-plus/blob/d9d0cc9063f85684179908569227dda251981751/src/inner_product_round.rs#L171-L178). In each case, the intent is to split each vector in half for folding operations.

A somewhat more idiomatic approach could be to perform a [split](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.split_at) directly.

This could also be a good opportunity to perform bounds checking in each inner-product round, instead of only [initially](https://github.com/tari-project/bulletproofs-plus/blob/d9d0cc9063f85684179908569227dda251981751/src/inner_product_round.rs#L86-L90), just in case the caller did something wonky between rounds.

AaronFeickert:
Currently, inner-product vectors are split in half using [pairs of slices](https://github.com/tari-project/bulletproofs-plus/blob/d9d0cc9063f85684179908569227dda251981751/src/inner_product_round.rs#L171-L178). Further, the consistency of vector lengths is checked [initially](https://github.com/tari-project/bulletproofs-plus/blob/d9d0cc9063f85684179908569227dda251981751/src/inner_product_round.rs#L86-L90), but not on each round, which relies on the caller to keep the vectors intact.

This PR makes three improvements. First, it replaces each slice pair with a [split](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.split_at), which reduces the likelihood of error and communicates the intent more clearly. Second, it limits the visibility of the inner-product functions to help guard against misuse. Finally, it adds a vector length consistency check to each round function as an internal assertion that the splits cannot panic.

Closes #55.

BREAKING CHANGE: Because this makes several functions non-public, it is technically a breaking change. However, there were no known use cases for this, meaning the change should be non-breaking in practice.

### DIFF
--- a/src/inner_product_round.rs
+++ b/src/inner_product_round.rs
@@ -31,7 +31,6 @@ pub struct InnerProductRound<'a, P> {
     h_base: P,
     y_powers: Vec<Scalar>,
     done: bool,
-    extension_degree: ExtensionDegree,
 
     // Prover data
     ai: Vec<Scalar>,
@@ -64,7 +63,7 @@ where
 {
     #![allow(clippy::too_many_arguments)]
     /// Initialize a new 'InnerProductRound' with sanity checks
-    pub fn init(
+    pub(crate) fn init(
         gi_base: Vec<P>,
         hi_base: Vec<P>,
         g_base: Vec<P>,
@@ -83,13 +82,13 @@ where
                 "Vectors gi_base, hi_base, ai, bi and y_powers cannot be empty".to_string(),
             ));
         }
-        if !(hi_base.len() == n && ai.len() == n && bi.len() == n) || (y_powers.len() != (n + 2)) {
+        if !(hi_base.len() == n && ai.len() == n && bi.len() == n && y_powers.len() == n + 2) {
             return Err(ProofError::InvalidArgument(
                 "Vector length for inner product round".to_string(),
             ));
         }
-        let extension_degree = ExtensionDegree::try_from_size(g_base.len())?;
-        if extension_degree as usize != alpha.len() {
+        let extension_degree = ExtensionDegree::try_from_size(g_base.len())? as usize;
+        if extension_degree != alpha.len() {
             return Err(ProofError::InvalidLength("Inconsistent extension degree".to_string()));
         }
         Ok(Self {
@@ -99,7 +98,6 @@ where
             h_base,
             y_powers,
             done: false,
-            extension_degree,
             ai,
             bi,
             alpha,
@@ -107,7 +105,7 @@ where
             b: None,
             r1: None,
             s1: None,
-            d1: Vec::with_capacity(extension_degree as usize),
+            d1: Vec::with_capacity(extension_degree),
             li: Vec::with_capacity(n * aggregation_factor + 2),
             ri: Vec::with_capacity(n * aggregation_factor + 2),
             transcript: transcript.into(),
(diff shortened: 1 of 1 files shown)


AaronFeickert:
This should be tested against the `tari-crypto` repository to ensure the visibility changes have no unintended side effects.

AaronFeickert:
Something went wonky and is holding up tests and benchmarks. Marking as draft until I figure out what went wrong.

src/inner_product_round.rs
```diff
@@ -31,7 +31,6 @@ pub struct InnerProductRound<'a, P> {
     h_base: P,
     y_powers: Vec<Scalar>,
     done: bool,
-    extension_degree: ExtensionDegree,
```
hansieodendaal, line 34: I would not remove `extension_degree` here

src/inner_product_round.rs
```diff
@@ -99,15 +98,14 @@ where
             h_base,
             y_powers,
             done: false,
-            extension_degree,
             ai,
```
hansieodendaal, line 101: Ditto `extension_degree`

src/inner_product_round.rs
```diff
@@ -117,9 +115,19 @@ where
...
-        let extension_degree = self.extension_degree as usize;
+        if !(self.hi_base.len() == n && self.ai.len() == n && self.bi.len() == n && self.y_powers.len() >= n + 2) {
+            return Err(ProofError::InvalidArgument(
+                "Vector length for inner product round".to_string(),
+            ));
+        }
+        let extension_degree = ExtensionDegree::try_from_size(self.g_base.len())? as usize;
+        if extension_degree != self.alpha.len() {
+            return Err(ProofError::InvalidLength("Inconsistent extension degree".to_string()));
+        }
```
hansieodendaal, line 129: I would keep the assignment of `extension_degree` in the constructor as it should never change
```suggestion
        let extension_degree = self.extension_degree as usize;
        if !(extension_degree == self.alpha.len() && extension_degree == self.g_base.len()) {
            return Err(ProofError::InvalidLength("Inconsistent extension degree".to_string()));
        }
```

src/inner_product_round.rs
```diff
@@ -167,15 +175,13 @@ where
...
-        let gi_base_hi = &self.gi_base[n..];
-        let hi_base_lo = &self.hi_base[..n];
-        let hi_base_hi = &self.hi_base[n..];
+        // Split vectors in half; since `n` is always a nontrivial power of two, this is well defined
+        n /= 2;
+        let (a1, a2) = self.ai.split_at(n);
+        let (b1, b2) = self.bi.split_at(n);
+        let (gi_base_lo, gi_base_hi) = self.gi_base.split_at(n);
+        let (hi_base_lo, hi_base_hi) = self.hi_base.split_at(n);
+
```
hansieodendaal, line 184: I do not see this as an improvement, especially looking at what goes on behind `pub const fn split_at(&self, mid: usize)` and the fact that it will panic if something is amiss, just the same as our code.

With the addition of the new check/verify we have these:
1. `let mut n = self.gi_base.len();`
2. `(self.hi_base.len() == n && self.ai.len() == n && self.bi.len() == n && self.y_powers.len() >= n + 2)`

That means the code as we have it is well protected and will never panic. That can even go as a comment above the slicing.

We can also reduce the lines of code if that is better:
```
let (a1, a2) = (&self.ai[..n], &self.ai[n..]);
etc.
```

src/inner_product_round.rs
```diff
@@ -117,9 +115,19 @@ where
...
-    pub fn inner_product<T: RngCore + CryptoRng>(&mut self, rng: &mut T) -> Result<(), ProofError> {
+    pub(crate) fn inner_product<T: RngCore + CryptoRng>(&mut self, rng: &mut T) -> Result<(), ProofError> {
+        // Ensure that vector lengths are still consistent, just in case
         let mut n = self.gi_base.len();
-        let extension_degree = self.extension_degree as usize;
+        if !(self.hi_base.len() == n && self.ai.len() == n && self.bi.len() == n && self.y_powers.len() >= n + 2) {
+            return Err(ProofError::InvalidArgument(
+                "Vector length for inner product round".to_string(),
+            ));
+        }
```
hansieodendaal, line 125: Good addition.

hansieodendaal (commented):
Please see the comments below.

src/inner_product_round.rs
AaronFeickert: Agreed that the panic condition doesn't change. The reason for the change was to make the intent (splitting the vectors) more clear, as well as to limit the possibility for error by only specifying the split index once per vector. The function `Vec::split_at` seemed perfect for this use case.

AaronFeickert:
Superseded by #57.

PR #57, opened 2023-08-07
feat: refactor inner-product prover

### LINKED ISSUE #55: Consider splitting inner-product vectors idiomatically

### LINKED ISSUE #61: Refactor inner-product prover
The inner-product prover is currently implemented as a standalone struct and round function. The range prover must instantiate the struct and iteratively perform the round function, which is annoying: it means asserting input consistency and some unnecessary moves.

We can simplify this by moving the same functionality directly in the range prover, replacing the iterative round function calls with a straightforward loop.

AaronFeickert:
The inner-product prover is a bit clunky, as it requires the range prover to set up a struct, perform iterative rounds, and then parse out the proof elements it needs. This also requires consistency checks that it would be nicer to avoid.

This PR moves the inner-product prover into the range prover, which simplifies things at the cost of a longer prover function.

It also fixes some [vector allocations](https://github.com/tari-project/bulletproofs-plus/blob/d9d0cc9063f85684179908569227dda251981751/src/inner_product_round.rs#L111-L112) that were much bigger than necessary.

Closes #61. Closes #55. Supersedes #56.


hansieodendaal (approved):
utACK

src/range_proof.rs
```diff
@@ -299,46 +302,183 @@ where
...
+
+        // Perform the inner-product folding rounds
+        while n > 1 {
+            n /= 2;
+
+            // Split the vectors for folding
+            let (a_lo, a_hi) = a_li.split_at(n);
+            let (b_lo, b_hi) = a_ri.split_at(n);
+            let (gi_base_lo, gi_base_hi) = gi_base.split_at(n);
+            let (hi_base_lo, hi_base_hi) = hi_base.split_at(n);
```
hansieodendaal, line 333:
```suggestion
            // Split the vectors for folding
            fn checked_split_at<T>(my_vec: &[T], n: usize) -> Result<(&[T], &[T]), ProofError> {
                if n <= my_vec.len() {
                    Ok(my_vec.split_at(n))
                } else {
                    Err(ProofError::InvalidLength("Index out of bounds".to_string()))
                }
            }
            let (a_lo, a_hi) = checked_split_at(&a_li, n)?;
            let (b_lo, b_hi) = checked_split_at(&a_ri, n)?;
            let (gi_base_lo, gi_base_hi) = checked_split_at(&gi_base, n)?;
            let (hi_base_lo, hi_base_hi) = checked_split_at(&hi_base, n)?;
```

hansieodendaal (commented):
Nice.

I would like one small change though to not hide the potential panic, so we can use a custom `fn checked_split_at<T>(my_vec: &[T], n: usize)` instead of `pub const fn split_at(&self, mid: usize)`

src/range_proof.rs
AaronFeickert: Doesn't this basically treat the split the same as before from a panic perspective? The slice-based version has the same issue.

AaronFeickert: Updated in a new commit.

hansieodendaal (approved):
ACK

SWvheerden (approved):

PR #58, opened 2023-08-09
docs: fix benchmark extension degrees

AaronFeickert:
A comment in the generator benchmarks describes how to use all supported extension degrees, but it used an outdated `enum`. This updates the comment so it is correct.


hansieodendaal (approved):
ACK

SWvheerden (approved):

PR #59, opened 2023-08-09
feat: compute verifier challenge sum more efficiently

AaronFeickert:
As part of proof verification, the verifier must compute the sum of consecutive nonzero powers of one of the Fiat-Shamir challenges. Currently, this is done naively by iteratively computing powers and adding them into an accumulator. Because the number of powers is equal to the product of the bit length and aggregation factor, this is nontrivial.

The computation can be made far more efficient. Because the sum of powers is a partial sum of a geometric series, it can be computed using [a well-known formula](https://mathworld.wolfram.com/GeometricSeries.html) (which is already used elsewhere in the verifier).

The change requires computation of an additional scalar inverse. As part of the optimization, all scalar inversions performed by the verifier are now included in a single batch operation per proof, which is more efficient.

Documentation is also updated to reflect the efficiency gain.

Overall, the result is an impressive improvement in verification. A single 64-bit range proof now verifies 7% faster.


hansieodendaal (approved):
ACK

SWvheerden (approved):

PR #63, opened 2023-08-10
fix: clean up getter functions and remove clones

AaronFeickert:
The getter functions for the inner-product round vectors have minor issues.

First, they perform an unnecessary cloning during decompression. Second, they return a `Result` that produces an error if the vectors are empty, which cannot happen in practice. Finally, the getters for the compressed vectors are entirely unnecessary.

This PR removes the cloning, gets rid of the compressed getters, and changes the return type of the decompressed getters. It also uses iterators where appropriate.

### DIFF
--- a/src/range_proof.rs
+++ b/src/range_proof.rs
@@ -902,7 +902,7 @@ where
                     "An item in member 'L' was not the canonical encoding of a point".to_string(),
                 ))
             })
-            .collect::<Result<Vec<_>, ProofError>>()
+            .collect()
     }
 
     // Helper function to decompress Ri
@@ -914,7 +914,7 @@ where
                     "An item in member 'L' was not the canonical encoding of a point".to_string(),
                 ))
             })
-            .collect::<Result<Vec<_>, ProofError>>()
+            .collect()
     }
 }
 
--- a/src/range_proof.rs
+++ b/src/range_proof.rs
@@ -895,28 +895,26 @@ where
 
     // Helper function to decompress Li
     fn li_decompressed(&self) -> Result<Vec<P>, ProofError> {
-        let mut li = Vec::with_capacity(self.li.len());
-        for item in &self.li {
-            li.push(item.decompress().ok_or_else(|| {
-                ProofError::InvalidArgument(
+        self.li
+            .iter()
+            .map(|p| {
+                p.decompress().ok_or(ProofError::InvalidArgument(
                     "An item in member 'L' was not the canonical encoding of a point".to_string(),
-                )
-            })?)
-        }
-        Ok(li)
+                ))
+            })
+            .collect::<Result<Vec<_>, ProofError>>()
     }
 
     // Helper function to decompress Ri
     fn ri_decompressed(&self) -> Result<Vec<P>, ProofError> {
-        let mut ri = Vec::with_capacity(self.ri.len());
-        for item in &self.ri {
-            ri.push(item.decompress().ok_or_else(|| {
-                ProofError::InvalidArgument(
-                    "An item in member 'R' was not the canonical encoding of a point".to_string(),
-                )
-            })?)
-        }
-        Ok(ri)
+        self.ri
+            .iter()
+            .map(|p| {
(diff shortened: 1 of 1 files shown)


src/range_proof.rs
```diff
@@ -751,54 +750,26 @@ where
 
     // Helper function to decompress Li
     fn li_decompressed(&self) -> Result<Vec<P>, ProofError> {
-        if self.li.is_empty() {
-            Err(ProofError::InvalidArgument("Vector 'L' not assigned yet".to_string()))
```
hansieodendaal, line 755: I would not remove this safety check

src/range_proof.rs
```diff
@@ -751,54 +750,26 @@ where
...
-                })?)
-            }
-            Ok(li)
-        }
-    }
-
-    // Helper function to return compressed Li
-    fn li(&self) -> Result<Vec<P::Compressed>, ProofError> {
-        if self.li.is_empty() {
-            Err(ProofError::InvalidArgument("Vector 'L' not assigned yet".to_string()))
```
hansieodendaal, line 772: ditto: I would not remove this safety check

src/range_proof.rs
```diff
@@ -751,54 +750,26 @@ where
...
+                    "An item in member 'L' was not the canonical encoding of a point".to_string(),
+                ))
+            })
+            .collect::<Result<Vec<_>, ProofError>>()
     }
 
     // Helper function to decompress Ri
     fn ri_decompressed(&self) -> Result<Vec<P>, ProofError> {
-        if self.ri.is_empty() {
-            Err(ProofError::InvalidArgument("Vector 'R' not assigned yet".to_string()))
```
hansieodendaal, line 781: ditto: I would not remove this safety check

src/range_proof.rs
```diff
@@ -751,54 +750,26 @@ where
...
-                })?)
-            }
-            Ok(ri)
-        }
-    }
-
-    // Helper function to return compressed Ri
-    fn ri(&self) -> Result<Vec<P::Compressed>, ProofError> {
-        if self.ri.is_empty() {
-            Err(ProofError::InvalidArgument("Vector 'R' not assigned yet".to_string()))
```
hansieodendaal, line 798: ditto: I would not remove this safety check

hansieodendaal (commented):
Nice use of iterators

Please see the comments below

src/range_proof.rs
AaronFeickert: It definitely won't be needed once #57 is merged and the inner-product prover is baked into the range prover.

Although to be honest, I don't think it's needed with the current setup either. Verification won't succeed on an empty vector anyway, and returning empty vectors seems like the correct generic behavior of the getters.

src/range_proof.rs
AaronFeickert: See above.

src/range_proof.rs
AaronFeickert: See above.

src/range_proof.rs
AaronFeickert: See above.

src/range_proof.rs
```diff
@@ -750,28 +750,26 @@ where
...
+            .map(|p| {
+                p.decompress().ok_or(ProofError::InvalidArgument(
                     "An item in member 'L' was not the canonical encoding of a point".to_string(),
-                )
-            })?)
-        }
-        Ok(li)
+                ))
+            })
+            .collect::<Result<Vec<_>, ProofError>>()
```
sdbondi, line 760: Not at all NB but specifying the types here is redundant because the return type already specifies the type you are collecting to, so a `.collect()` works fine.  Same for below.

hansieodendaal (approved):
ACK

src/range_proof.rs
AaronFeickert: Updated in a new commit!

AaronFeickert:
Looks like CI is currently broken due to an outdated nightly toolchain.

Since this is fixed by #68, this PR is blocked.

PR #70, opened 2023-08-25
feat!: generalize batch transcript labels

### LINKED ISSUE #64: Allow batch verifier to use separate transcript labels
Range proofs support the use of a caller-provided label that is used to initialize the proof transcript for context-based domain separation. However, batch verification [assumes a common label](https://github.com/tari-project/bulletproofs-plus/blob/d9d0cc9063f85684179908569227dda251981751/src/range_proof.rs#L432) among all proofs in a batch. This can be generalized to use a separate label for each batched proof.

AaronFeickert:
Currently, transcript labels are assumed to be [common](https://github.com/tari-project/bulletproofs-plus/blob/d9d0cc9063f85684179908569227dda251981751/src/range_proof.rs#L432) when verifying a batch of proofs. This PR generalizes the batch verifier API to accept a vector of labels, each of which is applied to the corresponding proof.

Closes #64.

BREAKING CHANGE: Changes the batch verification API.

### DIFF
--- a/benches/range_proof.rs
+++ b/benches/range_proof.rs
@@ -127,6 +127,7 @@ fn verify_aggregated_rangeproof_helper(bit_length: usize, extension_degree: Exte
             // 0. Batch data
             let mut statements = vec![];
             let mut proofs = vec![];
+            let mut transcript_labels = vec![];
 
             // 1. Generators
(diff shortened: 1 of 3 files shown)


hansieodendaal (approved):
ACK

SWvheerden (approved):

PR #73, opened 2023-08-27
test: remove test panic to improve coverage

AaronFeickert:
Removes a case where a test used `panic` in favor of an `is_err()` assertion, which is more idiomatic and improves coverage.


AaronFeickert:
The repository has very good test coverage, and we should keep it that way :)

hansieodendaal (approved):
ACK

hansieodendaal (approved):
ACK

SWvheerden (approved):

PR #74, opened 2023-08-31
ci: don't build dependency documentation

AaronFeickert:
Currently, CI builds documentation for all dependencies, which seems unnecessary.

Fortunately, there's a [flag](https://doc.rust-lang.org/cargo/commands/cargo-doc.html#documentation-options) to disable this behavior. This PR adds the flag to make CI a bit speedier.

PR #75, opened 2023-09-06
ci: update CI actions

AaronFeickert:
Unfortunately, the popular `actions-rs` project, which supported Rust CI, has been [unmaintained](https://github.com/actions-rs/toolchain/issues/216) for some time. In addition, the default repository `checkout` action has been [updated](https://github.com/actions/checkout/releases/tag/v4.0.0). Finally, documentation builds include dependencies, which slows down testing unnecessarily.

This PR addresses all of these issues.

First, it migrates from `actions-rs` to the [maintained](https://github.com/dtolnay/rust-toolchain) `dtolnay/rust-toolchain` action. Further, it updates the `checkout` action. Finally, it passes a [flag](https://doc.rust-lang.org/cargo/commands/cargo-doc.html#documentation-options) that ignores dependencies when building documentation. The result is a more modern and streamlined CI flow.

This supersedes #74.

### DIFF
--- a/.github/workflows/clippy-check.yml
+++ b/.github/workflows/clippy-check.yml
@@ -1,26 +1,25 @@
 on: [pull_request]
 name: Clippy
 jobs:
-  clippy_check:
+  format:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
-      - uses: actions-rs/toolchain@v1
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@master
         with:
-          components: clippy, rustfmt
+          components: rustfmt
           toolchain: nightly
-          override: true
-      - uses: actions-rs/cargo@v1
+      - name: Formatting
+        run: cargo +nightly fmt --all -- --check
+  clippy:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@master
         with:
-          command: fmt
-          args: --all -- --check
-      - name: Install cargo-lints
-        uses: actions-rs/cargo@v1
-        with:
-          command: install
-          args: cargo-lints
-      - name: Clippy lints
-        uses: actions-rs/cargo@v1
-        with:
-          command: lints
-          args: clippy --all-targets
+          components: clippy
+          toolchain: stable
+      - name: Install linter
+        run: cargo install cargo-lints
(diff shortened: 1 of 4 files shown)


hansieodendaal (commented):
We also need to run `cargo lints clippy  --all-targets`
Does `cargo clippy` result in `cargo clippy  --all-targets -- -D warnings`?

.github/workflows/clippy-check.yml
```diff
@@ -1,26 +1,23 @@
...
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@master
         with:
-          command: lints
-          args: clippy --all-targets
\ No newline at end of file
+          components: clippy
+          toolchain: stable
+      - name: Lints
+        run: cargo clippy --all-targets
```
stringhandler, line 23: this is not correct. The full command is 
`cargo lints clippy --all-targets` so that it uses the lints.toml file

.github/workflows/test.yml
```diff
@@ -13,29 +13,16 @@ jobs:
...
-        with:
-          command: check
-          args: --all-features --all-targets
+        run: cargo check --all-features --all-targets
       - name: test/debug
-        uses: actions-rs/cargo@v1
-        with:
-          command: test
-          args: --all-features
+        run: cargo test --all-features
```
stringhandler, line 24: We've found `nextest` very useful in debugging slow tests. Could be worth switching it over

AaronFeickert: Good idea! I'll test it out (pun intended).

.github/workflows/clippy-check.yml
AaronFeickert: Odd that `cargo clippy` won't use the lint file...

hansieodendaal: As far as I have it we need to run, or this is what I always run locally:
```
cargo clippy --all-targets --all-features -- -D warnings
```
and
```
cargo lints clippy --all-targets --all-features
```
Those are two separate checks.

AaronFeickert: I added the `clippy::all` group to the lints file as a deny, which should capture this behavior.

hansieodendaal (approved):
utACK

hansieodendaal (approved):
utACK

SWvheerden (approved):

PR #81, opened 2023-09-19
ci: deserialization fuzz testing

### LINKED ISSUE #80: Consider using `quickcheck` in CI
Work in #79 adds basic proof deserialization fuzzing. This is useful, but not build for use in CI. It was suggested to use [`quickcheck`](https://github.com/BurntSushi/quickcheck) to add similar CI functionality, which would be helpful as an automated sanity check against future changes that introduce deserialization bugs.

If feasible, this should be added to CI.

AaronFeickert:
This adds basic deserialization fuzz testing to CI via [`quickcheck`](https://github.com/BurntSushi/quickcheck). Because it uses a limited number of randomized test runs, it is not particularly thorough, and should not be considered a replacement for the fuzz testing added in #79. That being said, it's a reasonable sanity check and runs quickly.

Closes #80.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ rand_core = "0.6"
 [dev-dependencies]
 bincode = "1"
 criterion = "0.5"
+quickcheck = "1"
 
 [[bench]]
 name = "range_proof"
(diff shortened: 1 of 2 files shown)


SWvheerden (approved):

hansieodendaal:
Nice

PR #83, opened 2023-09-20
fix: add check for unused deserialization data

### LINKED ISSUE #82: Add extra check that all deserialization data is read
Serialized proof data is read as [chunks](https://github.com/tari-project/bulletproofs-plus/blob/e01c380186111c1fbe3fba213d1b492beff11b9d/src/range_proof.rs#L1017-L1020) to make parsing cleaner and safer. However, the function used can have [leftover data](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.chunks_exact) that does not yield a full chunk. This should never occur because of earlier length checks, but a separate [check](https://doc.rust-lang.org/stable/std/slice/struct.ChunksExact.html#method.remainder) should be added as an extra guard to ensure that serialized proofs are fully canonical.

AaronFeickert:
Adds an extra check for unused deserialization data to ensure proof serialization is canonical. Removes redundant length checks.

Currently, serialized proof data is read as [chunks](https://github.com/tari-project/bulletproofs-plus/blob/e01c380186111c1fbe3fba213d1b492beff11b9d/src/range_proof.rs#L1017-L1020) and [tuples](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.tuples) for parsing purposes. Each of these functions can have leftover data. However, existing length checks ensure such conditions cannot occur.

This PR replaces the length checks with a final check that there is no leftover data. This helps to assert that serialization is canonical and simplifies deserialization.

Closes #82.

### DIFF
--- a/src/range_proof.rs
+++ b/src/range_proof.rs
@@ -990,9 +990,9 @@ where
     P: Compressable,
     P::Compressed: FixedBytesRepr,
 {
-    /// Serializes the proof into a byte array
+    /// Serializes the proof into a canonical byte array
     /// The first byte is an encoding of the extension degree, which tells us the length of `d1`
-    /// Then we serialize the rest of the proof elements as 32-byte encodings
+    /// Then we serialize the rest of the proof elements as canonical byte encodings
     pub fn to_bytes(&self) -> Vec<u8> {
         // The total proof size: extension degree encoding, fixed elements, vectors
         let mut buf = Vec::with_capacity(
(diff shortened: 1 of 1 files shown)


AaronFeickert:
This should be carefully checked to ensure the use of `tuple()` on `chunks.by_ref()` is doing the right thing, and not setting it up so the `remainder()` call will always pass.

hansieodendaal (approved):
ACK

It seems to be watertight now

AaronFeickert:
At this point, we could even get rid of some of the initial length checks in favor of a simpler check that the inner-product folding vectors are not empty, which would be more straightforward and easier to test.

AaronFeickert:
@hansieodendaal This has been updated to remove redundant length checks.

hansieodendaal:
I would keep those length checks, as they are valid and add to the general understanding of the deserialization

hansieodendaal (approved):
ACK

CjS77 (approved):
LGTM

PR #87, opened 2023-10-09
fix: audit updates

AaronFeickert:
Makes several updates suggested from an audit. In particular, these relate to arithmetic operations and indexing. While none are tied to security issues, they are useful.


CjS77 (approved):
LGTM

PR #89, opened 2023-10-11
fix: prover cleanup

### LINKED ISSUE #88: Examine prover overflow risks
As noted in an audit, there are several places in the prover where overflow or underflow could occur. In practice, an honest prover wouldn't trigger these, meaning the practical risk is minimal (and a malicious prover can generate proofs however it wishes). Regardless, it's good practice to use checked operations where feasible in order to guard against future changes.

AaronFeickert:
Cleans up the prover by replacing some indexing with iterator operations to address suggestions made in an audit.

Closes #88.

### DIFF
--- a/src/range_proof.rs
+++ b/src/range_proof.rs
@@ -321,13 +321,13 @@ where
             *item -= z;
         }
         for (i, item) in a_ri.iter_mut().enumerate() {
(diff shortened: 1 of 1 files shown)


CjS77 (approved):
LGTM

Issue #86, opened 2023-09-25
Provide coverage reports for PRs


AaronFeickert:
Similarly to [this issue](https://github.com/tari-project/tari-crypto/issues/208), it would be useful to get coverage reports on PRs for easier review.

AaronFeickert:
After discussion with @CjS77, it sounds like there isn't a desire to implement this. Closing.

PR #90, opened 2023-10-11
chore!: unused method cleanup

AaronFeickert:
Removes public methods that became unused after #87 was merged.

BREAKING CHANGE: Removes unused methods that were public, which is technically a breaking change.


AaronFeickert:
Interestingly, this would have been caught by a fix to #86.

CjS77 (approved):

PR #92, opened 2023-10-13
ci: update nightly and fail on doc warnings

AaronFeickert:
Currently, CI testing uses a fixed nightly toolchain. While this is useful against a spurious nightly build, the repository already uses the latest nightly for formatting, lints, and checks. I think it's worth it to stick with the latest nightly across the board.

Additionally, this updates documentation checks to fail on warnings. This ensures that the documentation warnings don't go unnoticed.


CjS77 (approved):
LGTM

PR #95, opened 2023-10-21
chore: cleanup and index removal

AaronFeickert:
Removes more occurrences of vector indexing to reduce the risk of a panic. Does some additional cleanup. Adds a unit test to confirm correctness of one of the changes.

### DIFF
--- a/src/range_proof.rs
+++ b/src/range_proof.rs
@@ -848,11 +848,9 @@ where
             // Compute d's sum efficiently
             let mut d_sum = z_square;
             let mut d_sum_temp_z = z_square;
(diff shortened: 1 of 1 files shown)


hansieodendaal (approved):
Nice ( although not so nice to read now ;-) ), better for mainatainability

ACK

AaronFeickert:
> Nice ( although not so nice to read now ;-) ), better for mainatainability
> 
> ACK

Yeah, I briefly considered writing a wrapper for the `get`-and-map-error boilerplate, but that wouldn't be idiomatic.

SWvheerden (approved):

PR #98, opened 2023-10-23
fix: improve prover consistency checks

### LINKED ISSUE #97: Proof generation succeeds on invalid value
If the prover specifies a value that exceeds the allowed bit length, proof generation succeeds. This isn't a security issue, since the verifier will reject the proof. However, this change in behavior wasn't present at the last release, so it should be changed.

It may also be worth adding additional consistency checks that the provided prover witness are valid for a given statement, in order to be more consistent and thorough about what it means for the prover to return an error.

AaronFeickert:
When producing a proof, the prover does not currently check that the provided witness value is within the range specified by the statement bit length. It also does not check that the witness corresponds to valid openings of the statement commitments. This is not a security issue (the verifier will reject such an invalid proof), but it means that the caller may produce an invalid proof that does not return an error.

This PR adds these checks, and ensures that the prover returns an error if they fail. It also adds unit tests exercising various failure modes.

Closes #97.

### DIFF
--- a/src/range_proof.rs
+++ b/src/range_proof.rs
@@ -14,7 +14,7 @@ use std::{
 
 use curve25519_dalek::{
     scalar::Scalar,
-    traits::{Identity, IsIdentity, VartimePrecomputedMultiscalarMul},
+    traits::{Identity, IsIdentity, MultiscalarMul, VartimePrecomputedMultiscalarMul},
 };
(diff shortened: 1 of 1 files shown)


hansieodendaal (approved):
Nice

ACK

SWvheerden (approved):

PR #100, opened 2023-10-23
fix: don't panic on inconsistent generators

### LINKED ISSUE #99: Prover can panic on inconsistent input data
It's possible to cause the prover to panic by using a statement whose aggregation factor is inconsistent with its generators.

AaronFeickert:
Currently, the prover will panic if the statement aggregation factor does not precisely correspond to that of the inner-product generators. It would be best to allow the prover to use a larger generator set, similarly to how the verifier operates, in order to be more flexible. This PR adds such a fix, and includes a regression test.

Review of this PR should include asserting that removing the fix code will fail the regression test.

Closes #99.

### DIFF
--- a/src/range_proof.rs
+++ b/src/range_proof.rs
@@ -330,8 +330,10 @@ where
             }
         }
 
-        let mut gi_base: Vec<P> = statement.generators.gi_base_iter().cloned().collect();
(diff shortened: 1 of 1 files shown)


hansieodendaal (approved):
Nice one

ACK

SWvheerden (approved):

PR #103, opened 2023-11-29
feat: support `no_std`

### LINKED ISSUE #102: Make `std` optional if feasible
Proof operations in this library are complex when it comes to computation and memory use, but it may still be useful to make `std` an optional feature to ensure optimal flexibility for uses of the library. It's not immediately clear how feasible this would be, but it's worth investigating at some point.

AaronFeickert:
It may be useful for this library to play nicely with `no_std` targets, despite its rather heavy memory requirements. This PR ensures that the library is `no_std` compatible out of the box, without needing to fiddle with features.

Closes #102.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,13 +9,13 @@ description = "A smaller faster implementation of Bulletproofs"
 [dependencies]
 blake2 = { version = "0.10", default-features = false }
 byteorder = { version = "1", default-features = false }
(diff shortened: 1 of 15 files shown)


AaronFeickert:
Marking as draft since the library makes use of `rand` fairly extensively. This should be refactored such that callers must provide their own random number generators. We can still use `rand_core` to place marker trait restrictions on these generators to ensure they are cryptographically secure. This makes the public API a bit more annoying, but it's fairly common in the ecosystem.

AaronFeickert:
Superseded by #107.

PR #105, opened 2023-11-29
feat: simplify dependencies

### LINKED ISSUE #104: Refactor to make `rand` optional
The library currently makes use of `rand` fairly extensively for hard-coded random number generators (RNGs). This means `rand` is a dependency requirement, which hinders `no_std` support; see #102 for more on this. Further, it means the caller cannot specify an RNG of its choice, which reduces flexibility.

It's common in the ecosystem to allow the caller to provide its own RNG, despite making APIs marginally more irritating to use. We can still require cryptographically-secure RNGs using `rand_core` marker traits.

AaronFeickert:
Issue #104 recommends removing `rand` as a required dependency, as it hinders `no_std` support and reduces flexibility.

This PR updates dependencies in a few ways.

It introduces a default `rand` feature that requires `rand` as a dependency. If the `rand` feature is not enabled, the caller can instead call new `RangeProof::prove_with_rng` and `RangeProof::verify_batch_with_rng` functions that require the caller to provide its own cryptographically-secure random number generator whose quality is enforced by the `rand_core::CryptoRngCore` marker trait. Since the existing API is still present with the feature enabled, existing uses of the library shouldn't need to make any changes.

It updates CI to make sure the library builds without `rand` enabled. In this case, we don't run tests, since those require `rand`.

It also removes the unused `NonDebug` functionality, which simplifies the dependency tree.

Finally, it sets all other dependencies to their minimal feature set. This will be useful as part of an eventual move to support `no_std` targets.

Closes #104.


CjS77 (approved):
LGTM.

I believe there is a way to test the no-std portion of the code. We'd need a no-std rand lib and some tests that are `cfg(not(feature = "rand"))`.

PR #107, opened 2024-01-09
feat: support `no-std` environments

### LINKED ISSUE #102: Make `std` optional if feasible

AaronFeickert:
This PR makes the library friendly to `no-std` environments.

It also adds a new (default) `std` feature that enables corresponding functionality in dependencies.

It updates tests and benchmarks to use a seeded `ChaCha12`-based pseudorandom number generator to ensure these don't require `std` or `rand` (which makes CI much cleaner).

Finally, it adds a `wasm32-unknown-unknown` CI build target to ensure everything works as expected in a `no-std` environment.

Supersedes #103.

Closes #102.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,10 +12,10 @@ jobs:
       - name: toolchain
         uses: dtolnay/rust-toolchain@master
         with:
-          toolchain: stable
+          toolchain: nightly 
           targets: wasm32-unknown-unknown
       - name: build (WASM)
-        run: cargo build --target wasm32-unknown-unknown --no-default-features
+        run: cargo +nightly build --target wasm32-unknown-unknown --no-default-features -Zavoid-dev-deps
 
   test:
     name: cargo test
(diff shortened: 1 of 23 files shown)


SWvheerden (approved):
utACK

src/range_proof.rs
```diff
@@ -1256,26 +1253,6 @@ mod tests {
...
-        }
-
-        const TESTS: u64 = 100_000;
-
-        QuickCheck::new()
-            .min_tests_passed(TESTS)
-            .tests(TESTS)
-            .max_tests(TESTS)
-            .quickcheck(internal as fn(Vec<u8>) -> bool);
-    }
```
hansieodendaal, line 1277: I think we can keep this `QuickCheck` test

.github/workflows/test.yml
```diff
@@ -3,6 +3,20 @@ on: [push, pull_request]
 name: Test
```
hansieodendaal, line 3: Please add a test for `cargo test --release --no-default-features`

The doc test in `range_proof.rs` can be excluded from this test if appended like this:
```
/// #[cfg(feature = "rand")]
/// use rand::rngs::OsRng;
/// # fn main() {
/// #[cfg(feature = "rand")]
/// {
/// ...
```

hansieodendaal (changes requested):
Hi, looking good, just some small changes please.

hansieodendaal (approved):
ACK

PR #109, opened 2024-01-09
feat: use `TranscriptRng` for random number generation

### LINKED ISSUE #108: Use `TranscriptRng` for prover nonces
Because the library uses Merlin for transcript operations, it's possible to use its `TranscriptRng` functionality to make prover nonces more robust. This uses a combination of the transcript state, witness data, and an external cryptographically-secure random number generator (RNG) to ensure that even in case of a suboptimal RNG, effective entropy falls back to that of the witness.

This functionality should be added.

AaronFeickert:
As noted in #108, it's possible to improve the robustness of prover nonces by using Merlin's `TranscriptRng` functionality.

This PR moves all prover nonce generation to this design. In particular, it ensures that the most up-to-date state of the transcript is used each time the prover needs secure randomness (per the Merlin [documentation](https://docs.rs/merlin/latest/merlin/struct.TranscriptRngBuilder.html#note)). The verifier also uses this to produce its random weights.

It also takes the opportunity to refactor transcript operations so they are cleaner and likely safer to use.

Closes #108.

### DIFF
--- a/src/range_proof.rs
+++ b/src/range_proof.rs
@@ -271,8 +271,8 @@ where
             }
         }
 
-        // Start a new transcript and get an RNG
-        let mut transcript = RangeProofTranscript::<P>::new(
+        // Start a new transcript and generate the transcript RNG
+        let (mut transcript, mut transcript_rng) = RangeProofTranscript::<P>::new(
             transcript_label,
             &statement.generators.h_base().compress(),
             statement.generators.g_bases_compressed(),
@@ -280,8 +280,9 @@ where
             extension_degree,
             aggregation_factor,
             statement,
+            Some(witness),
+            rng,
         )?;
-        let mut transcript_rng = transcript.build_rng(witness, rng);
 
         // Set bit arrays
         let mut a_li = Zeroizing::new(Vec::with_capacity(full_length));
@@ -324,9 +325,8 @@ where
             statement.generators.g_bases().iter(),
         );
 
-        // Update transcript, get challenges, and generate RNG
-        let (y, z) = transcript.challenges_y_z(&a.compress())?;
-        transcript_rng = transcript.build_rng(witness, rng);
+        // Update transcript, get challenges, and update RNG
+        let (y, z) = transcript.challenges_y_z(&mut transcript_rng, rng, &a.compress())?;
 
         let z_square = z * z;
 
@@ -464,8 +464,10 @@ where
                     .chain(hi_base_hi),
             ));
 
-            // Update transcript, get challenge, and generate RNG
+            // Update transcript, get challenge, and update RNG
             let e = transcript.challenge_round_e(
+                &mut transcript_rng,
+                rng,
                 &li.last()
                     .ok_or(ProofError::InvalidLength("Bad inner product vector length".to_string()))?
                     .compress(),
@@ -473,7 +475,6 @@ where
                     .ok_or(ProofError::InvalidLength("Bad inner product vector length".to_string()))?
                     .compress(),
             )?;
-            transcript_rng = transcript.build_rng(witness, rng);
             let e_square = e * e;
             let e_inverse = e.invert();
             let e_inverse_square = e_inverse * e_inverse;
@@ -551,8 +552,8 @@ where
             b += g_base * eta;
         }
 
-        // Update transcript and get challenge (no RNG needed)
-        let e = transcript.challenge_final_e(&a1.compress(), &b.compress())?;
+        // Update transcript, get challenge, and update RNG
+        let e = transcript.challenge_final_e(&mut transcript_rng, rng, &a1.compress(), &b.compress())?;
         let e_square = e * e;
 
         let r1 = *r + a_li[0] * e;
@@ -823,11 +824,8 @@ where
                 return Err(ProofError::InvalidLength("Vector L/R length not adequate".to_string()));
             }
 
-            // Batch weight (may not be equal to a zero valued scalar) - this may not be zero ever
-            let weight = Scalar::random_not_zero(rng);
-
             // Start the transcript
-            let mut transcript = RangeProofTranscript::new(
+            let (mut transcript, mut transcript_rng) = RangeProofTranscript::new(
                 transcript_label,
                 &h_base_compressed,
                 g_bases_compressed,
@@ -835,17 +833,22 @@ where
                 extension_degree,
                 aggregation_factor,
                 statement,
+                None,
+                rng,
             )?;
 
             // Reconstruct challenges
-            let (y, z) = transcript.challenges_y_z(&proof.a)?;
+            let (y, z) = transcript.challenges_y_z(&mut transcript_rng, rng, &proof.a)?;
             let challenges = proof
                 .li
                 .iter()
                 .zip(proof.ri.iter())
-                .map(|(l, r)| transcript.challenge_round_e(l, r))
+                .map(|(l, r)| transcript.challenge_round_e(&mut transcript_rng, rng, l, r))
                 .collect::<Result<Vec<Scalar>, ProofError>>()?;
-            let e = transcript.challenge_final_e(&proof.a1, &proof.b)?;
+            let e = transcript.challenge_final_e(&mut transcript_rng, rng, &proof.a1, &proof.b)?;
+
+            // Batch weight (may not be equal to a zero valued scalar) - this may not be zero ever
+            let weight = Scalar::random_not_zero(&mut transcript_rng);
 
             // Compute challenge inverses in a batch
             let mut challenges_inv = challenges.clone();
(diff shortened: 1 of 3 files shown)


sdbondi (commented):
LGTM

One optimisation question: how do you interpret [this note](https://docs.rs/merlin/latest/merlin/struct.TranscriptRngBuilder.html#note) in the docs? Since the witness data cannot be updated within the  `prove_with_rng` function (immutable reference), does the TranscriptRng need to be reinitialized multiple times within the function? It's not clear to me if they mean the `TranscriptRng` should be kept up to date with the witness data or if each time it is used once to generate a random number a new instance is needed (seems doubtful/doesn't make sense, but if that is the case this is not done everywhere e.g. lines 524,525 etc).

src/range_witness.rs
```diff
@@ -37,6 +37,21 @@ impl RangeWitness {
             extension_degree: extension_degree.try_into()?,
         })
     }
+
+    /// Produce a byte representation of the witness data for use in transcript RNG seeding
+    /// Note that this does _not_ need to be canonical!
+    pub(crate) fn to_bytes(&self) -> Zeroizing<Vec<u8>> {
+        let mut bytes = Zeroizing::new(Vec::<u8>::new());
```
sdbondi, line 44: nit(perf): could perform a single allocation (though it is a bit of a pain :smile: )

```suggestion
        let size = self.openings.iter().map(|o| size_of::<u64>() + o.r.len() * 32).sum();
        let mut bytes = Zeroizing::new(Vec::<u8>::with_capacity(size));
```

AaronFeickert:
> Since the witness data cannot be updated within the `prove_with_rng` function (immutable reference), does the TranscriptRng need to be reinitialized multiple times within the function?

Unfortunately yes. It's built from a clone of the current `Transcript` state, which is updated each round. This means reinitialization is necessary to take advantage of those state changes.

> It's not clear to me if they mean the `TranscriptRng` should be kept up to date with the witness data or if each time it is used once to generate a random number a new instance is needed (seems doubtful/doesn't make sense, but if that is the case this is not done everywhere e.g. lines 524,525 etc).

The witness data is fixed at the start of proof generation and doesn't change. The only reason it needs to be added to each `TranscriptRng` reinitialization is because it isn't (and can't be) part of the transcript state since it's known only to the prover. It makes it a bit awkward to use, but I get why they designed the API that way.

The linked documentation is a bit unclear when it advises reinitializing the `TranscriptRng` every time the protocol needs randomness. Because the transcript state is only updated _between_ rounds, there is no advantage to reinitializing _within_ a round. That's the approach I took: the `TranscriptRng` is reinitialized whenever the transcript state is updated, and used for all randomness needed during that round.

That being said, I could have made the intent more clear by including better comments and placing the reinitializations more directly at the state changes. I'll make a small update to do so, and to include your other review suggestion.

AaronFeickert:
I made more changes than originally expected in order to make transcript and random number generator functionality more clear.

In particular, the previous transcript helper functions have been moved into `RangeProofTranscript` to abstract away the underlying Merlin transcript from the prover and verifier. I also moved `RangeWitness::to_bytes` out of `RangeWitness` and into `RangeProofTranscript` since the resulting byte representation is not canonical and should never directly be used outside of this specific use case.

The prover and verifier transcript operations were correspondingly simplified, and each `TranscriptRng` initialization now appears alongside a transcript state update.

sdbondi (approved):
Thanks, LGTM - code is much clearer

hansieodendaal (changes requested):
Hi looking good, just some small changes, please.

AaronFeickert:
After discussing with @hansieodendaal, it was agreed that it would be best to move the transcript RNG updates into `RangeProofTranscript` to ensure they aren't forgotten.

The most recent commit does this. Transcript initialization returns a `RangeProofTranscript` and a `TranscriptRng`; calls to the challenge functions update the `TranscriptRng` via a mutable reference.

The verifier now also uses a `TranscriptRng` for its random weights. This allows it to use the same API as the prover, except without passing a witness as secret data.

src/range_proof.rs
```diff
@@ -276,18 +271,17 @@ where
...
+        let mut transcript = RangeProofTranscript::<P>::new(
+            transcript_label,
             &statement.generators.h_base().compress(),
             statement.generators.g_bases_compressed(),
             bit_length,
             extension_degree,
             aggregation_factor,
             statement,
         )?;
+        let mut transcript_rng = transcript.build_rng(witness, rng);
```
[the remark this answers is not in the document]
AaronFeickert: This is now part of the `RangeProofTranscript` API.

src/range_proof.rs
```diff
@@ -330,8 +324,10 @@ where
             statement.generators.g_bases().iter(),
         );
 
-        // Get challenges
-        let (y, z) = transcripts::transcript_point_a_challenges_y_z(&mut transcript, &a.compress())?;
+        // Update transcript, get challenges, and generate RNG
+        let (y, z) = transcript.challenges_y_z(&a.compress())?;
+        transcript_rng = transcript.build_rng(witness, rng);
```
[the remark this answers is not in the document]
AaronFeickert: This is now part of the `RangeProofTranscript` API.

src/range_proof.rs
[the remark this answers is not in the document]
AaronFeickert: This is now part of the `RangeProofTranscript` API.

src/transcripts.rs
```diff
@@ -1,76 +1,116 @@
...
+            .iter()
+            .map(|o| size_of::<u64>() + o.r.len() * size_of::<Scalar>())
+            .sum();
+        let mut witness_bytes = Zeroizing::new(Vec::<u8>::with_capacity(size));
+        for opening in &witness.openings {
+            witness_bytes.extend(opening.v.to_le_bytes());
+            for r in &opening.r {
+                witness_bytes.extend(r.as_bytes());
+            }
+        }
```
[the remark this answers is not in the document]
AaronFeickert: This is now stored in the `RangeProofTranscript`.

src/transcripts.rs
```diff
@@ -1,76 +1,199 @@
...
+                witness_bytes.extend(opening.v.to_le_bytes());
+                for r in &opening.r {
+                    witness_bytes.extend(r.as_bytes());
+                }
+            }
+
+            Some(witness_bytes)
+        } else {
+            None
+        };
```
sdbondi, line 108: nit: (optional) 

```rust
impl RangeWitness {

pub(super) fn to_non_canonical_bytes(&self) -> Zeroize<Vec<u8>> {
    let size: usize = witness
                .openings
                .iter()
                .map(|o| size_of::<u64>() + o.r.len() * size_of::<Scalar>())
                .sum();
            let mut witness_bytes = Zeroizing::new(Vec::<u8>::with_capacity(size));
            for opening in &witness.openings {
                witness_bytes.extend(opening.v.to_le_bytes());
                for r in &opening.r {
                    witness_bytes.extend(r.as_bytes());
                }
            }

            witness_bytes
}
}
```

```suggestion
        let witness_bytes = witness.map(|w| w.to_non_canonical_bytes())
```

src/transcripts.rs
```diff
@@ -1,76 +1,199 @@
...
+        // Set up the RNG
+        let transcript_rng = Self::build_rng(&transcript, bytes.as_ref(), external_rng);
+
+        Ok((
+            Self {
+                transcript,
+                bytes,
+                _phantom: PhantomData,
+            },
+            transcript_rng,
```
sdbondi, line 119: (optional) Could we not encapsulate the transcript_rng in the `Transcript`?

For scalar generation, we could either expose the RNG
```rust
    pub(crate) fn rng_mut(&mut self) -> &mut TranscriptRng {
        &mut self.transcript_rng
    }
```
or perhaps

```rust
pub(crate) fn generate_random_scalar(&mut self) -> Scalar {
   Scalar::random_not_zero(&mut transcript.rng)
}
```

sdbondi (approved):
LGTM

hansieodendaal (approved):
Awesome

ACK

src/transcripts.rs
AaronFeickert: That's more or less what I had in an initial version. I took the byte serialization out of `RangeWitness` since it's not suitable (and possibly not safe) to use it in other contexts. Because it's so tightly linked to the transcript use case, it seemed safer to lock it away in `RangeProofTranscript` instead.

Interestingly, we could just make the serialization canonical if we really wanted to; it wouldn't matter from a transcript perspective. But that seems unnecessary since serializing secret data isn't something that should be done without a good reason.

src/transcripts.rs
AaronFeickert: Ooh, I like the first idea, since it keeps everything more nicely contained. I was even hoping to have `RangeProofTranscript` hold a mutable reference to the external RNG so the prover and verifier wouldn't be able to misuse it, but the lifetimes got messy and I decided against it.

There should be some sanity tests for the resulting `TranscriptRng`, so I might add this suggestion in a follow-up PR.

PR #111, opened 2024-01-12
feat: move RNGs into `RangeProofTranscript`

AaronFeickert:
Recent work in #109 uses Merlin's `TranscriptRng` functionality to improve handling of random number generation, and also refactors transcript operations for safer use through a `RangeProofTranscript` wrapper.

A [suggestion](https://github.com/tari-project/bulletproofs-plus/pull/109#discussion_r1449861417) by @sdbondi recommends moving the `TranscriptRng` into `RangeProofTranscript` so the prover and verifier aren't responsible for it. This PR adds such a change.

It also updates `RangeProofTranscript` to hold a mutable reference to the external random number generator. This ensures that the prover and verifier can't accidentally use it instead of the `TranscriptRng`.

### DIFF
--- a/src/transcripts.rs
+++ b/src/transcripts.rs
@@ -28,12 +28,9 @@ use crate::{
 ///
 /// When the prover initializes the wrapper, it includes the witness as the secret data.
 /// The verifier doesn't have any secret data to include, so it passes `None` instead.
-/// In either case, you get a `RangeProofTranscript` and a `TranscriptRng`.
+/// In either case, you get a `RangeProofTranscript` that is updated when you use the challenge functions.
 ///
-/// When the transcript is updated using the challenge functions, you must provide the `TranscriptRng`, which is also
-/// updated.
-///
-/// When randomness is needed, just use the `TranscriptRng`.
+/// When you need randomness, use `as_mut_rng()` to get an `&mut TranscriptRng` for this purpose.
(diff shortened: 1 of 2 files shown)


AaronFeickert:
Marking as draft to add some tests.

src/transcripts.rs
```diff
@@ -36,26 +36,30 @@ use crate::{
...
-pub(crate) struct RangeProofTranscript<P>
+pub(crate) struct RangeProofTranscript<'a, P, R>
 where
     P: Compressable + Precomputable,
     P::Compressed: FixedBytesRepr + IsIdentity,
+    R: CryptoRngCore,
 {
     transcript: Transcript,
     bytes: Option<Zeroizing<Vec<u8>>>,
+    rng: TranscriptRng,
```
hansieodendaal, line 47: I think the previous naming is better suited to the intent. It is also still in line with the comment `/// When randomness is needed, just use the `TranscriptRng`.`
```suggestion
    transcript_rng: TranscriptRng,
```

hansieodendaal (approved):
ACK

src/transcripts.rs
AaronFeickert: Updated.

AaronFeickert:
Do not merge; I just noticed that the transcript documentation is out of date.

sdbondi (approved):
Looks great!

PR #113, opened 2024-02-07
ci: move to stable toolchain

AaronFeickert:
This PR moves all CI to the stable toolchain, with the single exception of formatting (which requires nightly). It does this by cleaning up the `rand` feature to use `rand_core::OsRng`, which lets the `no_std` build run properly on stable.

Unfortunately, there is no good way to do this while keeping `quickcheck` around. This was only used for very limited fuzz testing; while having any fuzz testing be part of CI is nice, its practical benefit is likely minimal. Further, its functionality is available through the non-CI fuzzer.

### DIFF
--- a/.github/workflows/clippy-check.yml
+++ b/.github/workflows/clippy-check.yml
@@ -22,4 +22,4 @@ jobs:
       - name: Install linter
         run: cargo install cargo-lints
       - name: Lints
-        run: cargo +stable lints clippy --all-targets
+        run: cargo +stable lints clippy --all-targets --all-features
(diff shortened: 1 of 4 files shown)


hansieodendaal (commented):
We could hide the `QuickCheck` fuzzing behind a feature flag and only do CI for that using nightly (similar to what was requested in [monero-rs](https://github.com/monero-rs/monero-rs)). I find it quite handy to have that kind of testing as part of CI.

AaronFeickert:
I think that requires nightly, since without a nightly-only compiler flag, all development dependencies are pulled in. This breaks on targets that don't support `getrandom` (a `quickcheck` requirement).

hansieodendaal (approved):
utACK

PR #118, opened 2024-02-13
chore: arithmetic lint

### LINKED ISSUE #117: Add arithmetic side effect lint
The [Quarkslab audit report](https://github.com/tari-project/bulletproofs-plus/blob/da71f7872f02a0e9d3000c316bb083181daa9942/docs/quarkslab-audit/README.md) used the `clippy::arithmetic_side_effects` lint to check for operations that could overflow, underflow, or otherwise be undefined. This lint wasn't added to the CI list since it was not clear how to use its [configuration](https://rust-lang.github.io/rust-clippy/master/index.html#/arithmetic_side_effects) to avoid triggering on Ristretto operations that are always well defined.

After some investigation, it looks like the `clippy.toml` syntax is very picky:
```
arithmetic-side-effects-allowed = [ "tari_curve25519_dalek::Scalar" ]
```

This doesn't appear to work for `RistrettoPoint`, presumably because of the use of the generic `P` used throughout the codebase. This syntax works for it:
```
arithmetic-side-effects-allowed = [ "P" ]
```
However, this seems risky since it depends on this specific naming. Fortunately, only a small number of operations are caught by this, and can be manually allowed.

AaronFeickert:
The [Quarkslab audit](https://github.com/tari-project/bulletproofs-plus/blob/da71f7872f02a0e9d3000c316bb083181daa9942/docs/quarkslab-audit/README.md) of this library recommended the use of an arithmetic side-effect lint. This was not done at the time due to difficulty getting the lint to work with Ristretto types, which do not pose risk.

This PR adds the lint and makes necessary changes to account for it.

Closes #117.

### DIFF
--- a/benches/range_proof.rs
+++ b/benches/range_proof.rs
@@ -38,6 +38,7 @@ static EXTRACT_MASKS: [VerifyAction; 1] = [VerifyAction::VerifyOnly];
 // static EXTENSION_DEGREE: [ExtensionDegree; 3] = [ExtensionDegree::Zero, ExtensionDegree::Two, ExtensionDegree::Four];
 // static EXTRACT_MASKS: [VerifyAction; 2] =  [VerifyAction::VerifyOnly, VerifyAction::RecoverOnly];
 
+#[allow(clippy::arithmetic_side_effects)]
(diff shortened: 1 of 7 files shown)


hansieodendaal (approved):
ACK

PR #120, opened 2024-03-05
fix: pin nightly version on source cov workflow

CjS77:
A bandaid over the issue of source coverage failing due to Dalek being a naughty library again.

Clearly a temporary fix, since this will break again some thim in the future.


AaronFeickert:
Can stable be used here?

AaronFeickert (approved):
Doesn't look like this can be tested without merging, but LGTM.

PR #124, opened 2024-03-05
ci: run more CI in parallel

AaronFeickert:
Splits out more CI jobs to run in parallel. This is an experiment to see if we get noticeably faster runs.


hansieodendaal:
Maybe you can add some stats in the PR description? That way it would be easier to make an objective assessment.

SWvheerden (approved):

PR #127, opened 2024-03-22
fix: reduce verification vector allocation

### LINKED ISSUE #126: Verification vectors are over-allocated
The dynamic scalar and point vectors used in batch verification are allocated in part based on the aggregation factor associated to each statement's parameters. This aggregation factor does not specify the number of commitments associated to the statement, but rather the _maximum_ number of supported commitments, as specified by the number of associated generators.

As a result, it's possible for a statement to specify a larger aggregation factor than is actually used; this allows for more flexibility and efficiency when producing generators. The result is that the verification vectors may be over-allocated.

The fix is to use the number of commitments associated to a statement instead of the specified aggregation factor.

AaronFeickert:
Batch verification vectors are allocated in part using the expected aggregation factor of each statement in the batch. However, this was being done using the aggregation factors corresponding to each statement's associated generators, which may exceed the aggregation factors actually used in the statements. The result was a possible over-allocation of these vectors.

This PR changes the allocation to use the actual aggregation factors.

Closes #126.

### DIFF
--- a/src/range_proof.rs
+++ b/src/range_proof.rs
@@ -778,7 +778,7 @@ where
         let mut msm_dynamic_len = extension_degree.checked_add(1).ok_or(ProofError::SizeOverflow)?;
         for (statement, proof) in statements.iter().zip(range_proofs.iter()) {
             msm_dynamic_len = msm_dynamic_len
-                .checked_add(statement.generators.aggregation_factor())
+                .checked_add(statement.commitments.len())
                 .ok_or(ProofError::SizeOverflow)?;
             msm_dynamic_len = msm_dynamic_len.checked_add(3).ok_or(ProofError::SizeOverflow)?;
             msm_dynamic_len = msm_dynamic_len


AaronFeickert:
One way to test this PR is to check (just before the final verification multiscalar multiplication evaluation) that `dynamic_scalars.len() == msm_dynamic_len && dynamic_points.len() == msm_dynamic_len` after the change, but that this check fails before the change.

sdbondi (approved):

sdbondi:
The reasoning and the code make sense. Is it correct to assume that in Tari's usage, the aggregation factor and number of commitments are always the same?

hansieodendaal (approved):
utACK

hansieodendaal:
> The reasoning and the code make sense. Is it correct to assume that in Tari's usage, the aggregation factor and number of commitments are always the same?

Yes they are , see https://github.com/tari-project/bulletproofs-plus/pull/127/files#diff-a4997ec3ef089e27ef54a31d22f718c1e73a454188bab2ec2f96d55f172aa687L239

AaronFeickert:
@sdbondi yes, but not quite for the reason that @hansieodendaal gave.

There's an unfortunate use of terminology: parameters specify an aggregation factor that represents the _largest_ number of commitments that can be accommodated, but within the proof mechanics we also specify the aggregation factor to be the _actual_ number of commitments handled by any given proof. This design was intended to be flexible in cases where the caller needs to verify proofs where the aggregation factor is not fixed, but is merely capped.

It is the case that the Tari codebase currently only generates parameters that account for trivial (non)-aggregation, so the case handled by this PR wouldn't arise. But in the more general case where the parameters can accommodate more commitments than are provided to the verifier, it would arise and result in a (small) over-allocation.

AaronFeickert:
It might be beneficial to do some renaming, such that the aggregation factor specified in parameters is called `maximum_aggregation_factor` or something, in order to better identify what it does.

PR #129, opened 2024-03-22
feat: remove partial precomputation

### LINKED ISSUE #93: Update curve library dependency
It was recommended in an audit to use a more up-to-date fork of the curve library, since [upstream](https://github.com/dalek-cryptography/curve25519-dalek) finalized its latest major release. It should be straightforward to rebase the [fork](https://github.com/tari-project/curve25519-dalek) against this.

### LINKED ISSUE #96: Replace scalar exponentiation after curve library update
There are a couple of places in the codebase where scalar exponentiation is required. This is done efficiently using the standard square-and-multiply approach, but it is done manually.

If the curve library dependency is updated to support its upstream's `group` feature, exponentiation will be [directly supported](https://docs.rs/curve25519-dalek/latest/curve25519_dalek/scalar/struct.Scalar.html#method.pow_vartime) and should be used instead if feasible.

Related to #93.

### LINKED ISSUE #128: See if we can remove partial precomputation
Currently, the only reason to require a custom curve library fork is to support partial precomputation. This permits a more flexible use of inner-product generators.

While it's beneficial to have this flexibility, it's a headache for integration with other libraries. It might be useful to revisit this design choice and see if it can be removed in favor of being able to use the upstream curve library directly. This has become increasingly relevant since upstream has seen a flurry of activity that continues to diverge from the fork.

For use cases in the Tari ecosystem, it's unlikely to cause efficiency problems in practice. For use cases elsewhere that do require a larger inner-product generator set than is used in verification, there are options. Either the caller generates parameters for the aggregation factors it actually needs (which incurs no internal efficiency hit), or we allow for a larger generator set but pad the corresponding scalars (which does incur an efficiency hit).

AaronFeickert:
Partial precomputation is the sole reason for maintaining a custom curve library fork, which has proven to be a headache and limits compatibility with other libraries.

This PR removes partial precomputation altogether. If you supply parameters with more inner-product generators than you need, padding is used. This incurs an efficiency hit, but only in this particular case. For the use cases in the Tari ecosystem, this is not an issue.

As a result of this change, we also switch back to the latest version of the upstream curve library and simplify some scalar exponentiation operations. The audit report is also updated to note the curve library dependency change.

Closes #128. Closes #93. Closes #96.

### DIFF
--- a/src/range_proof.rs
+++ b/src/range_proof.rs
@@ -8,7 +8,7 @@
 use alloc::{string::ToString, vec, vec::Vec};
 use core::{
     convert::{TryFrom, TryInto},
-    iter::once,
+    iter::{once, repeat},
     marker::PhantomData,
     ops::{Add, Mul, Shr},
     slice::ChunksExact,
@@ -330,8 +330,19 @@ where
                 Scalar::random_not_zero(range_proof_transcript.as_mut_rng())
             });
         }
+        let padding = 2usize
+            .checked_mul(statement.generators.bit_length())
+            .ok_or(ProofError::SizeOverflow)?
+            .checked_mul(statement.generators.aggregation_factor())
+            .ok_or(ProofError::SizeOverflow)?
+            .checked_sub(a_li.len())
+            .ok_or(ProofError::SizeOverflow)?
+            .checked_sub(a_ri.len())
+            .ok_or(ProofError::SizeOverflow)?;
         let a = statement.generators.precomp().vartime_mixed_multiscalar_mul(
-            a_li.iter().interleave(a_ri.iter()),
+            a_li.iter()
+                .interleave(a_ri.iter())
+                .chain(repeat(&Scalar::ZERO).take(padding)),
             alpha.iter(),
             statement.generators.g_bases().iter(),
         );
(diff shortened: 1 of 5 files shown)


AaronFeickert:
Note that this PR was written such that if partial precomputation is added upstream, it will be easy to support by effectively reverting the changes here.

src/range_proof.rs
```diff
@@ -1023,8 +1035,20 @@ where
...
         // Perform the final check using precomputation
+        let padding = 2usize
+            .checked_mul(max_statement.generators.bit_length())
+            .ok_or(ProofError::SizeOverflow)?
+            .checked_mul(max_statement.generators.aggregation_factor())
+            .ok_or(ProofError::SizeOverflow)?
+            .checked_sub(max_mn)
+            .ok_or(ProofError::SizeOverflow)?
+            .checked_sub(max_mn)
+            .ok_or(ProofError::SizeOverflow)?;
```
hansieodendaal, line 1046: I think this and the previous use^ can go into a function
```rust
fn calculate_padding<P: FromUniformBytes + Compressable + Clone + Precomputable>(
    generators: &RangeParameters<P>,
    a_li_len: usize,
    a_ri_len: usize,
) -> Result<usize, ProofError> {
    2usize
        .checked_mul(generators.bit_length())
        .ok_or(ProofError::SizeOverflow)?
        .checked_mul(generators.aggregation_factor())
        .ok_or(ProofError::SizeOverflow)?
        .checked_sub(a_li_len)
        .ok_or(ProofError::SizeOverflow)?
        .checked_sub(a_ri_len)
        .ok_or(ProofError::SizeOverflow)
}
```

hansieodendaal (commented):
Looks good. Maybe we can add a unit test for the case where _If you supply parameters with more inner-product generators than you need, padding is used._?

src/range_proof.rs
AaronFeickert: Done.

AaronFeickert:
Updated to ensure that at least version `4.1.1` of `curve25519-dalek` is used. This is required to support the `group` feature that lets us use built-in efficient scalar exponentiation.

hansieodendaal (approved):
ACK

AaronFeickert:
Rebased, with some commit cleanup.

AaronFeickert:
Partial precomputation is [now supported](https://github.com/dalek-cryptography/curve25519-dalek/pull/668) in the curve library.

PR #132, opened 2024-04-01
chore: rename for maximum aggregation factor

### LINKED ISSUE #130: Use better naming for parameter aggregation factor
Parameters contain an aggregation factor that represents the maximum number of commitments that can be handled in a single proof. However, the codebase also uses "aggregation factor" to refer to the actual number of commitments a proof contains. This already caused minor confusion that necessitated #127, and suggests the naming should be changed.

Specifically, parameters should probably specify that their aggregation factor is merely a cap.

AaronFeickert:
The aggregation factor used in range parameters refers more specifically to the _maximum_ aggregation factor supported by the parameters. This PR does some renaming to make this more clear.

Closes #130.


sdbondi (approved):
LGTM definitely clearer

PR #133, opened 2024-04-01
test: check that the serialization constant is correct

AaronFeickert:
This PR adds a test that the serialized element size constant is correct for Ristretto.


hansieodendaal (approved):
utACK

PR #140, opened 2024-08-08
feat: proof failure blame

AaronFeickert:
This PR adds blame functionality to proof batch verification, offering either single or full failure identification. In each case, the verifier first attempts to verify a complete batch. Subsequent behavior depends on the caller's requirements.

After initial failure using `RangeProof::verify_batch_with_first_blame`, the verifier performs a binary search to identify the first failing proof in the batch, providing its index. This requires a number of batch verifications that is logarithmic in the batch size; each subsequent verification cuts the number of proofs in the batch by half.

After initial failure using `RangeProof::verify_batch_with_full_blame`, the verifier attempts to verify each proof in the batch, providing a list of indexes of all failing proofs. This requires a number of non-batch verifications that is linear in the batch size.

This supersedes tari-project/tari-crypto#224.


AaronFeickert:
This still needs tests.

PR #142, opened 2024-09-03
chore: remove use of default for points and scalars

### LINKED ISSUE #141: Replace use of point and scalar defaults
The codebase sometimes uses the `Default` trait to get either the identity point of the Ristretto group (in either compressed or uncompressed form) or the zero scalar. This is fine in practice due to the implementation of that trait on the relevant types. However, it seems non-idiomatic and has a bad code smell.

It would be more idiomatic to use the `Identity` trait implementation for Ristretto types and the `ZERO` constant for scalars. This makes much more unambiguous what the intent is.

AaronFeickert:
The use of the `Default` implementation for Ristretto point and scalar types is somewhat non-idiomatic, since it isn't immediately clear what a "default" point or scalar must be. In practice, these are the identity group element and zero scalar, respectively.

This PR removes ambiguity and makes the library more future proof by removing all such uses. Instead, it uses the `Identity` implementation for point types and the `ZERO` constant for scalars.

Closes #141.

### DIFF
--- a/src/protocols/transcript_protocol.rs
+++ b/src/protocols/transcript_protocol.rs
@@ -80,7 +80,7 @@ impl TranscriptProtocol for Transcript {
 
 #[cfg(test)]
 mod test {
-    use curve25519_dalek::RistrettoPoint;
+    use curve25519_dalek::{traits::Identity, RistrettoPoint};
     use merlin::Transcript;
 
     use super::*;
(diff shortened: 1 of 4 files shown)


hansieodendaal (approved):
utACK

AaronFeickert:
Any other changes needed to merge?

AaronFeickert:
Is anything blocking a merge?

SWvheerden (approved):

PR #151, opened 2026-07-15
perf: speed up batch verification

sdbondi:
## Summary

Speeds up batch verification without changing what is accepted or rejected. Medians on an M-series MacBook, 64-bit proofs (harness included as `examples/verify_bench.rs`):

| Case | main | this PR | Change |
|---|---|---|---|
| 1 proof | 0.73 ms | 0.69 ms | −5% |
| 16 proofs | 4.90 ms | 4.34 ms | −11% |
| 64 proofs | 18.7 ms | 13.7 ms | −27% |
| 256 proofs | 77.0 ms | 50.3 ms | −35% |
| 64 proofs, `RecoverOnly` | 8.8 ms | 3.1 ms | −65% |
| 64 proofs, `RecoverAndVerify` | 20.5 ms | 14.8 ms | −28% |

## Changes

1. **Choose the final multiscalar multiplication strategy by estimated cost.** The final check always used the precomputed-Straus mixed MSM, whose per-point cost is constant (~51 curve additions per dynamic point). Each proof contributes ~16 dynamic points, so large batches ran far past the size where Pippenger (~33 additions per point, and falling with size) overtakes it. Verification now estimates both costs, mirroring the algorithm parameters in `curve25519-dalek`, and picks the cheaper; small batches keep the precomputed path. Both strategies compute the same sum, so acceptance is unchanged.

2. **Skip redundant generator consistency comparisons.** The batch consistency check compared generator vectors element-wise between statements — tens of thousands of point-equality checks for large batches of statements cloned from the same `RangeParameters`. The precomputation table `Arc` is created exactly once alongside the generator vectors it is built from, so `Arc::ptr_eq` proves the vectors identical; mismatches still fall back to the full element-wise comparison.

3. **Halve the scalar work in the per-element loop.** The batch weight is folded into per-proof constants, and `d[i] * y^(nm−i)` is tracked incrementally (one multiplication per element, with a correction at aggregation-block boundaries) instead of materializing the `d` vector — roughly 10 → 5 scalar multiplications per element, and one fewer O(mn) allocation per proof.

4. **`RecoverOnly` skips verification-only work.** Point decompression (3 + 2·rounds points per proof) ran before mask recovery even though recovery never touches the points. Decompression now happens after the `RecoverOnly` continue, which is where the ~2.8× recovery speedup comes from. Behavior note: `RecoverOnly` no longer rejects proofs whose points are non-canonical encodings, since it no longer decompresses them; the structural length checks still run, and both verifying modes are unchanged.

5. **Mask recovery does one inversion per proof** instead of two per proof per extension degree, by algebraically folding the `e²` and `z²·y^(nm+1)` divisions into `mask = (d1 − η − e·d − residue·e²) · (e²·z²·y^(nm+1))⁻¹`.

6. **Removed per-proof clones** of `commitments`, `d1`, and `minimum_value_promises`.

## Testing

- Full test suite passes; `cargo lints clippy --all-targets --all-features` is clean; the `no_std` build (`--no-default-features`) checks; nightly `cargo fmt --check` is clean.
- New test `test_large_batch_invalid_proof_is_rejected` exercises rejection through the plain-MSM path (a 100-proof batch crosses the dispatch threshold).
- Reproduce the numbers with `cargo run --release --example verify_bench`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/range_proof.rs
```diff
@@ -972,10 +945,32 @@ where
                 },
             }
 
-            // Aggregate the generator scalars
-            let mut y_inv_i = Scalar::ONE;
-            let mut y_nm_i = y_nm;
+            // Everything hereafter is only needed for verification, so mask-only recovery skips it entirely,
+            // including the point decompressions
 
+            // Decompress the proof elements
```
sdbondi: **RecoverOnly no longer validates proof point encodings** (behavior change, `Err` → `Ok`)

Decompression of `a`/`a1`/`b`/`li`/`ri` moved after the `RecoverOnly` `continue`, so `verify_batch(..., VerifyAction::RecoverOnly)` now returns `Ok(masks)` for proofs whose points are non-canonical encodings. On `main` these returned `ProofError::InvalidArgument`. A caller recovering masks from stored/attacker-supplied proof bytes silently loses that well-formedness check.

src/range_proof.rs
```diff
@@ -1045,21 +1060,63 @@ where
...
+            } else if size < 800 {
+                (7, 64)
+            } else {
+                (8, 128)
+            };
+            #[allow(clippy::arithmetic_side_effects)]
+            let columns = 256 / window + 1;
+            columns.saturating_mul(size.saturating_add(buckets)).saturating_add(256)
+        };
+        let batch_is_valid = if pippenger_cost < straus_cost {
```
sdbondi: **Pippenger branch drops the aggregation-size overflow guard**

This branch skips `compute_generator_padding`, which was the only verify-time guard against commitments exceeding `max_aggregation_factor` — and curve25519-dalek's zip-based MSM silently truncates mismatched iterator lengths. `RangeStatement.commitments` is a `pub Vec`, so a caller can grow it past the limit after construction: on `main` that returned `SizeOverflow`; here, when the Pippenger path is taken, `gi_base_iter().take(max_mn)` yields fewer points than scalars, the zip truncates and misaligns every subsequent dynamic point, and the caller gets a misleading `VerificationFailed("Range proof batch not valid")` instead of the size error.

src/range_parameters.rs
```diff
@@ -111,6 +111,15 @@ where P: FromUniformBytes + Compressable + Clone + Precomputable
     pub fn precomp(&self) -> Arc<P::Precomputation> {
         self.bp_gens.precomp.clone()
     }
+
+    /// Returns `true` if `self` and `other` are known to share the same underlying generator data.
+    ///
+    /// The precomputation table is only ever created in `BulletproofGens::new`, alongside the generator vectors it is
+    /// built from, and cloning shares it; so pointer equality of the tables guarantees the generator vectors are
+    /// identical without an element-wise comparison. A `false` result is inconclusive.
+    pub(crate) fn shares_generator_data(&self, other: &Self) -> bool {
```
sdbondi, line 117: **`Arc::ptr_eq` shortcut rests on an unenforced invariant**

This shortcut skips the batch generator-consistency comparison based on the invariant that precomp is only created in `BulletproofGens::new` — but that invariant lives only in this comment, while `g_vec`/`h_vec` remain `pub(crate)` and mutable after construction. Crate-internal code that clones gens (sharing the `Arc`) and then mutates the clone's vectors — the same mutation pattern the tests at `range_proof.rs:1654`/`1676` use, just on fresh gens today — would make `shares_generator_data` report identical generators without comparing them, letting mismatched Gi/Hi pass `verify_statements_and_generators_consistency`. Worth enforcing type-level (private fields) or falling back to the full comparison.

sdbondi (commented):
Review of the batch-verification speedup. The refactored algebra checks out (mask-recovery folding, incremental `d_y` tracker, weight foldings, generator ordering across both MSM branches), and the full release-mode test suite passes — including the 257-proof batch that takes the Pippenger path and still rejects a tampered proof. Five findings inline, most severe first.

src/range_proof.rs
```diff
@@ -1045,21 +1060,63 @@ where
...
+        //
+        // Evaluation against the precomputed tables (Straus) avoids computing lookup tables for the static
+        // generators, but its per-point cost is constant, whereas Pippenger's per-point cost shrinks as the problem
+        // grows. For large batches Pippenger wins even though it must process the static generators from scratch.
+        // The estimates below count curve operations, mirroring the algorithms in `curve25519-dalek`:
+        // - precomputed Straus: 256 doublings, ~256/9 additions per static scalar (width-8 NAF), and table construction
+        //   (~8 operations) plus ~256/6 additions (width-5 NAF) per dynamic point
+        // - Pippenger: ~256/w + 1 columns, each costing an addition per point plus half a bucket set, with the window
+        //   width `w` chosen by problem size
+        let static_count = max_mn.saturating_mul(2);
```
sdbondi, line 1109: **MSM cost model hardcodes curve25519-dalek-internal tuning constants**

The Straus-vs-Pippenger crossover estimate bakes in dalek's NAF densities (29/51), window thresholds (500/800, widths 6/7/8), and 256-bit scalars — in code generic over any `P: Precomputable`, with no test or shared definition tying the numbers to the library they mirror. A dalek upgrade that retunes window selection (or a non-Ristretto backend) silently invalidates the crossover and picks the slower MSM path on the hottest path in the crate. Consider extracting this into a named, unit-testable helper or a hook on `Precomputable` so the backend owns its cost model.

examples/verify_bench.rs
```diff
@@ -0,0 +1,143 @@
...
+
+struct Case {
+    label: String,
+    transcripts: Vec<Transcript>,
+    statements: Vec<RangeStatement<curve25519_dalek::ristretto::RistrettoPoint>>,
+    proofs: Vec<RistrettoRangeProof>,
+    action: VerifyAction,
+}
+
+fn build_case(batch_size: usize, aggregation_factor: usize, action: VerifyAction, rng: &mut ChaCha12Rng) -> Case {
```
sdbondi: **Duplicates the existing criterion bench harness**

This 143-line example copies the case construction from `benches/range_proof.rs` nearly line-for-line (same seed 8675309, same ChaCha12Rng, same `minimum_values = value/3`, same seed_nonce pattern) while hand-rolling weaker warmup/median timing than criterion provides. `verify_batched_rangeproofs_helper` / `verify_aggregated_rangeproof_helper` already bench `verify_batch` across batch sizes and aggregation factors — extending their parameter lists would cover these cases without a second harness to keep in sync.

src/range_proof.rs
sdbondi: Fixed in ba8bb9b: decompression restored before the RecoverOnly early-exit, with a regression test (test_recover_only_rejects_invalid_point_encodings).

src/range_proof.rs
sdbondi: Fixed in ba8bb9b: compute_generator_padding hoisted above the MSM strategy choice so both paths return SizeOverflow, with a regression test (test_batch_rejects_commitments_exceeding_generator_capacity).

src/range_parameters.rs
sdbondi: Fixed in ba8bb9b: g_vec/h_vec/precomp are now private to BulletproofGens with test-only accessors, so the invariant is enforced by field privacy; the ptr_eq shortcut moved next to the fields as shares_precomp.

src/range_proof.rs
sdbondi: Fixed in ba8bb9b: extracted to estimate_straus_cost/estimate_pippenger_cost with docs pinning the constants to curve25519-dalek 5.0.0-pre.6 and a unit test of the crossover (test_msm_cost_estimates).

examples/verify_bench.rs
sdbondi: Fixed in ba8bb9b: example removed; the commented-out full-spectrum bench constants now include RecoverAndVerify.

SWvheerden (approved):

